PR #47, opened 2019-03-17
update to beta.5; by hand due to travis cred issues

dhalbert:
`build_board_info.py` was unable to run with adabot credentials, so I ran it by hand and did this PR by hand. It's equivalent to the automated PR. Could someone review and approve to get beta.5 into circuitpython.org? Thanks.

### DIFF
--- a/_data/files.json
+++ b/_data/files.json
@@ -6,35 +6,35 @@
             {
                 "files": {
                     "ID": [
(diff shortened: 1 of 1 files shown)


tannewt:
I may have broken the credentials when I deleted a token I showed on my livestream on Friday. Will look tomorrow. Thanks!

PR #108, opened 2019-03-31
add battery charger to feather_m4_express.md

TG-Techie:


ptorrone:
this was added via the features that are now in advanced search thank you @TG-Techie 

adafruit_2019_1771

PR #115, opened 2019-04-04
Stats Page

jwcooper:
Unlisted stats page on /stats. Uses github api to parse and populate the stats. Rework a bit of the css to share it a bit better.

Fixes: https://github.com/adafruit/circuitpython-org/issues/98

### DIFF
--- a/assets/css/main.scss
+++ b/assets/css/main.scss
@@ -15,3 +15,4 @@
 @import 'pages/downloads';
 @import 'pages/download';
 @import 'pages/awesome';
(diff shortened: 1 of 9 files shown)


sommersoft:
> We could have adabot create a stats json file to show here for libraries.

Should be doable. Framework is already there for Bundle downloads, and individual lib downloads from PyPI. [Example](https://adafruit-circuit-python.s3.amazonaws.com/adabot/bin/adabot/library_download_stats_20190405.txt)

PR #127, opened 2019-04-11
adding two boards (sam32 and kicksat-sprite)

maholli:
SAM32 board is [already on the master CP branch](https://github.com/adafruit/circuitpython/tree/master/ports/atmel-samd/boards/sam32), kicksat-sprite will be added shortly.

- pictures for both boards will be updated shortly


tannewt:
I believe @ptorrone added these already.

PR #152, opened 2019-04-15
Create trademarks.html

ptorrone:
@jwcooper i think i did this the correct way, if i did not lemme know, i'll fix and then continue to make a few other odds and ends pages we'll link to later...

### DIFF
--- a/trademarks.html
+++ b/trademarks.html
@@ -1,0 +1,13 @@
+---
+layout: home
+title: Trademarks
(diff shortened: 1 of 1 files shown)


jwcooper:
That looks good to me. Do you want to merge this now, or wait until the full text is in?

ptorrone:
thanks! @jwcooper deploy away! i'll do a another PR later for the text and graphics.

PR #158, opened 2019-04-17
Blinka pages

jwcooper:
I think we could leave this unlisted until we build up the content.

The "download instructions" button links to whatever is set in the front matter on the board.

Two new required fields for blinka:

```
download_instructions: "https://learn.adafruit.com/circuitpython-on-raspberrypi-linux/installing-circuitpython-on-raspberry-pi"
blinka: true
```
Screenshots:

circuitpython.org/blinka:
Blinka

circuitpython.org/blinka/raspberry_pi
Raspberry_Pi_Download

Let me know if you have any suggestions or changes.

cc @pt


jwcooper:
Good idea, removed the sort by download option.

Issue #119, opened 2019-04-06
LIBRARIES page


kattni:
Here is what I would like to see for the LIBRARIES page. 

Main page would be three sections:
* CircuitPython Libraries
    * (quick intro to libraries)
* Download Instructions (heading): 
    * Embedded links to the latest bundles.
    * Instructions on copying individual libraries to CIRCUITPY drive.
* Contributing (heading):
    * Quick bit about contributing to the libraries
    * New Libraries list (updated by script)
    * Updated libraries list (updated by script)
    * Link to Contributing page.

Contributing page:
Contributing page linked from the LIBRARIES main page will be information aggregated from all the libraries, updated by a script. It will be text blocks interspersed with lists of links to PRs/Issues/Repos (supplied by the script). The formatting will involve section headings, text and links. I would provide text content.

The script:
The script would supply the assets for the lists on the main page and for building the Contributing page. The script will be written by @sommersoft. It will be designed to update the New/Updated libraries lists, and the Contributing page, on a to-be-determined time interval, likely starting with weekly. We would add a cron job to Travis for circuitpython-org to run the script.

For both the lists on the main page, and the content on the Contributing page, the script can run and update a json file which can be used to populate the pages.

Terrible mock-up included.

![LIBRARIES_mock_up](https://user-images.githubusercontent.com/7865090/55662726-341d3480-57e4-11e9-8a24-c8382f78ffe8.png)

kattni:
@jwcooper @tylerdcooper I'd like work together with you to move forward with this as soon as is reasonable. I understand there may be other priorities.

jwcooper:
@kattni I can start on it this week, I hope.

For the script component (if possible), I'd like to follow the same process we use for the files.json (https://github.com/adafruit/circuitpython-org/blob/master/_data/files.json) circuitpython release process. For libraries, we could simply add a _data/libraries.json file that is pushed to this repository on each release. Then, we can just pull in that data in jekyll, etc.

sommersoft:
@jwcooper,

Using the same process as `files.json` shouldn't be an issue. The idea was to have the script in adabot anyway. It will most likely run in a Travis cron job, vs the release builds on circuitpython. Submitting PRs can be the same...

I'll try and get a schema together soon, though it honestly may just come straight out of GitHub's API replies.

sommersoft:
@jwcooper,

How does the below schema work for you all?

_Truncated Raw Example:_
```
[
  {
    "library_updates": {
[...]

PR #174, opened 2019-04-28
Add adabot For Auto-updating Libraries/Contrib Info

sommersoft:
The next (last?) part to #119.

Two environment variables will need to be added to Travis:
- `ADABOT_GITHUB_ACCESS_TOKEN`: self explanatory
- `CP_ORG_UPDATER_RUN_DAY`: Sets the day of the week that the script will actually run. Day is a number between 0 & 6, with 0 being Monday. (Travis cron will be set to run daily. This ensures the script runs on the day we want since Travis won't guarantee that a weekly cron will run on a certain day.)

I was unsure that the `rvm` stuff in the original `.travis.yml` was necessary, since the `script` section is empty. If its not needed I can remove it. Also, I just thought about this: I could also exclude it from the daily cron, using the `$TRAVIS_EVENT_TYPE` env var if need be.

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,3 +1,6 @@
 [submodule "awesome-circuitpython"]
 	path = awesome-circuitpython
 	url = https://github.com/adafruit/awesome-circuitpython
+[submodule "adabot"]
+	path = adabot
+	url = https://github.com/adafruit/adabot.git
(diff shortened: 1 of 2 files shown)


jwcooper:
You can remove the unused bits. I think rvm is not needed there, likely a remnant from originally setting up the site.

PR #191, opened 2019-05-21
Remove ugame10.md

deshipu:
Since the firmware built for µGame no longer works, it doesn't make
sense to offer it for download here -- it only makes people confused.


tannewt:
Why remove it when it'll work with the release next week?

tannewt:
ugame10 should be fixed with 4.0.1 so I'm closing this.

deshipu:
Thank you for fixing it.

PR #218, opened 2019-06-04
Resize and optimize image sizes

jwcooper:
WIP Yet...

Script to resize originally uploaded files to two different sizes, optimize sizes and load them properly for viewport size.

Current /downloads size: 29.23MB
New /downloads size: 1.47MB

To start, we'll either need people to upload 3 images for new boards (original, 300 px width, 700 px width), or run the script after adding a new board. Updated readme for the instructions on adding 3 images.


tannewt:
Is it possible to fall back to the scaled original if it is only available?

jwcooper:
We can do it with javascript, but I haven't found a way with only html/css that works reliably. I'll add the javascript fallback before we're ready to review.

tannewt:
Could we use Jekyll to do it?

jwcooper:
Great idea, will try that quick. Looks like you can test if a file exists with jekyll.

jwcooper:
We'll now fallback to the original image if the large or small sizes do not exist (using jekyll).

PR #255, opened 2019-07-28
Robo HAT MM1 M0 & M4

wallarug:
In this pull request:
 - added robohat m4
 - updated robohat m0 to reflect name changes in circuitpython

### DIFF
--- a/_board/robohatmm1_m0.md
+++ b/_board/robohatmm1_m0.md
@@ -1,8 +1,8 @@
 ---
 layout: download
-board_id: "robohatmm1"
+board_id: "robohatmm1_m0"
 title: "Robo HAT MM1 Download"
-name: "Robo HAT MM1"
+name: "Robo HAT MM1 M0"
 manufacturer: "Robotics Masters"
 board_url: "https://www.crowdsupply.com/robotics-masters/robo-hat-mm1"
 board_image: "robohatmm1.jpg"
--- a/_board/robohatmm1_m4.md
+++ b/_board/robohatmm1_m4.md
@@ -1,0 +1,26 @@
+---
+layout: download
+board_id: "robohatmm1_m4"
+title: "Robo HAT MM1 Download"
+name: "Robo HAT MM1 M4"
+manufacturer: "Robotics Masters"
+board_url: "https://www.crowdsupply.com/robotics-masters/robo-hat-mm1"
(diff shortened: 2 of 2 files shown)


_board/robohatmm1_m0.md
```diff
@@ -1,8 +1,8 @@
 ---
 layout: download
-board_id: "robohatmm1"
+board_id: "robohatmm1_m0"
```
tannewt: This will break downloads for the older versions. Would you rather keep the m0 name the same and just add the m4?

_board/robohatmm1_m4.md
```diff
@@ -0,0 +1,26 @@
...
+  - Robotics
+---
+
+The Robotics Masters Robo HAT MM1 is an open source robotics controller board for Raspberry Pi. It is education focused but works in many applications. The Robo HAT provides all the hardware you need in one simple, easy-to-use form factor. It removes the initial barriers to starting any robotics project.
+
+With support for Adafruit CircuitPython and other libraries the Robo HAT is able to act as a single solution for all projects great and small.
+
+The Robo HAT MM1 removes the need for buying lots of individual components by including them on the board. It can act as a PWM driver, provide on-board power through voltage regulators, and measure movement with the on-board IMU - all at the same time.
+
+The wide compatibility with many open source software platforms such as CircuitPython means you can use all existing software for sensors while creating your project in the way that works best for you. The hardware schematics and software are fully open source.
```
tannewt: I'd suggest describing what makes the M4 version different here.

_board/robohatmm1_m0.md
wallarug: It's ok to change.  Once PR #2014 gets merged in this won't break.  So this is on hold until I can fix up #2014 from adafruit/circuitpython

_board/robohatmm1_m4.md
wallarug: Done!  At a hardware level - not much is actually different.  They both use the same PCB layout.  Only difference is the processor.  I have added in some notes as per your recommendation.

_board/robohatmm1_m0.md
wallarug: PR 2014 has now been merged in.  So we should be good to go without breaking anything with this change.

wallarug:
Ready.  Thank you for feedback @tannewt .

PR #256, opened 2019-07-30
Add information about snekboard

keith-packard:
Snekboard is designed to control Lego robots using Power Functions
motors, servos and switches along with whatever other devices you can
devise. Snekboard runs CircuitPython and Snek.

Signed-off-by: Keith Packard <keithp@keithp.com>


ladyada:
🐍 🐍 🐍 🐍

PR #267, opened 2019-09-02
Updates to links and picture

ITACAInnovation:
Updated the link to the store where it is now possible to buy uChip and the picture with the final revision.


ladyada:
looks like there's a conflict - once its fixed we can merge :)

ITACAInnovation:
> 
> 
> looks like there's a conflict - once its fixed we can merge :)

Thanks for the support!
I noticed that the folder structure for the pictures changed since the last commit we did.
Should  we change our branch accordingly or is someone else already proceeding?
Thanks

ladyada:
you'll need to change your branch/commit

PR #269, opened 2019-09-04
Remote Libraries Data

jwcooper:
Use remote libraries data stored in s3 so daily updates of the libraries file can be performed. Will need to build site using GitHub Actions as there is a new plugin that is not supported by the default github-pages builds.

Prior to merging this:
- [x] Enable GitHub Actions for daily builds at a specified time
- [x] Move libraries.json automated builds to the adafruit-circuit-python s3 bucket. Currently have it located here: adafruit-circuit-python/adabot/web/libraries.json

The site will now be built by GitHub actions. Whenever there are changes on master, the workflow will build and commit the site to the gh-pages branch. It will also auto-build at 10am every day (need to confirm the timezone of github actions, but presuming it's utc). We can adjust the schedule build times to whatever works best.

The action is using an adabot personal access token. Adabot is the only user allowed to commit to the gh-pages branch.

Once this is merged, we need to modify the github pages source to use the gh-pages branch.


kattni:
We're working on getting the libraries.json automated builds to the S3 bucket. Thanks for getting this PR set up!

PR #290, opened 2019-09-18
For better or for worse, circuitpython.org doesn't have an RSS/Atom feed

jepler:
htmlproofer diagnosed this problem:
```
- ./_site/404/index.html
  *  External link http://localhost:4000/feed.xml failed: 404 No error
```


ptorrone:
would very much like a rss feed.

PR #292, opened 2019-09-18
Added STM32F407 board

makermelissa:


ladyada:
this doenst have a build yet but will!

makermelissa:
Yeah, I made an educated guess as to what it will be called. :)

PR #313, opened 2019-10-23
Only Run the GitHub Actions Workflow On the Adafruit Repository

sommersoft:
Ever since the GitHub Actions workflow to build the site was put in, the cron job has also been running on my fork (I'm a beta participant in Actions).

Obviously, it fails since I don't have the secrets, which triggers a daily email about the failure. Its a minor nuisance, but once Actions availability increases this might get out of hand.

Unfortunately, the way that `schedule` events currently work, it cannot be limited itself. I tried on a rest repo.

This change will only run the `build` job if the repository is owned by Adafruit. I added a job to print out if the repository's owner is Adafruit; otherwise its just a blank completed run that constantly shows "Starting your workflow runs...".

### DIFF
--- a/.github/workflows/workflow.yml
+++ b/.github/workflows/workflow.yml
@@ -8,8 +8,23 @@ on:
     - cron: 0 10 * * *
 
 jobs:
+  check-repo-owner:
+    # This job is so the entire workflow will end successfully and give some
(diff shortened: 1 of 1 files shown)


jwcooper:
Great idea, thank you!

PR #321, opened 2019-11-04
Adds images and markdown for the SparkFun Qwiic Micro

edspark:
Adds three images - small, large, and original sizes for the SparkFun Qwiic Micro. Also adds a markdown file with its information.


ladyada:
great work - merged :)

PR #337, opened 2019-12-12
Updated ShIRtty board photo

makermelissa:
Author provided a better photo.


ladyada:
do you need all 3 sizes?

makermelissa:
Unless something has changed, that my understanding was that all 3 were supposed to be provided. The large one is used when you are viewing the board details, the small one is used on the screen that displays all the boards and the original is supposed to be used as a fallback if the other images aren't available.

makermelissa:
We could probably get away with adding the large and small only at this point in time. It would be nice if there were automatic resizing and caching of the resized images.

ladyada:
oh wait i see the diff doesnt remove the files - it looked like they were deleted...all good!

PR #339, opened 2019-12-12
Removed a couple more missed features

makermelissa:


makermelissa:
If you'd like Microphone to be there, there were several boards with this.

PR #346, opened 2019-12-20
Added Bluebird and Blackpill Boards

makermelissa:
This should take care of those unknown boards.


ladyada:
thanks @makermelissa !

PR #380, opened 2020-01-17
fix feature list for open book

joeycastillo:
This was an oversight on my part; I didn't realize that the feature list was intended for tags and not freeform text. 

We could also add all the defined tags to template.md with a comment, which would clarify this for folks submitting new boards.


makermelissa:
Thanks for the template suggestion, I'll need to look into that. Would you mind opening an issue with that suggestion?

Issue #376, opened 2020-01-16
Add recently added Orange Pi boards


yeyeto2788:
In order to keep track of the changes, I'm creating this issue so after PR in Blinka is merged, all boards added in Blinka should be also added here.

P.D: @makermelissa we can do same procedure as in Blinka for this issue (assign it to me) I'll take care after boards are added.

makermelissa:
Fixed via #383

PR #383, opened 2020-01-19
Add Orange Pi Boards (One, Lite, Plus 2E)

yeyeto2788:
This solve issue #376 

Still missing small and large images for Orange Pi One and Lite, I'll be adding them tomorrow or Monday.


yeyeto2788:
Hey! how can I add notes on the page? Is there any format to follow? Is just to mention that the pinout on Orange Pi One and Lite the pins are shifted 180 degrees.

makermelissa:
You could probably just put it in the description section, but what do you mean that they're shifted 180 degrees? According to pics and pinout diagrams I could find, it appears they are the same direction.

yeyeto2788:
Hey @makermelissa! 

That's what I thought at first but as for https://linux-sunxi.org/Xunlong_Orange_Pi_One_%26_Lite if you head over the expansion port it states that shifting 180 degrees.

I can add the same info as in sunxi in the description as you suggested.

Regards.

yeyeto2788:
Regarding the images, I commented when opening the PR that I'll add them I just didn't get the time to edit them :( sorry for that. I'll do it tomorrow for sure.

I did add the original ones and it just missing the small and large right? Or you ment that even the original are wrong?

makermelissa:
Oh, gotcha. I missed that. If the small and large are present, the original dimensions don't really matter. It's supposed to be a fallback image, but using one large image that's loaded can slow things down.

yeyeto2788:
No problem, I'll do it tomorrow for sure so all images are ok and site doesn't take long to load, thank you very much for your comments. I really appreciate you taking the time on reviewing my endless PRs hahaha.

makermelissa:
I see what you mean, they're shifted 180 degrees from the Raspberry Pi rather than from each other. It probably doesn't matter to add it to here. This is more the type of thing that would be put into a guide.

yeyeto2788:
Ok, so I'll fix the images and we are ok, right? If you want me to do something regarding the guide just let me know.

Also if there is anything else you want me to change in this PR just tell me.

_blinka/orange_pi_lite.md
```diff
@@ -0,0 +1,51 @@
...
+downloads_display: true
+blinka: true
+features:
+  - Wi-Fi
+  - Bluetooth/BLE
+  - HDMI
+  - 40-pin GPIO
+---
+
+What’s Orange Pi PC？It’s an open-source single-board computer. It can run Android 4.4, Ubuntu, Debian Image. It uses the AllWinner H3 SoC, and has 512MB DDR3 SDRAM.
```
makermelissa: We should probably change this to Orange Pi Lite. Just grab the info from http://www.orangepi.org/orangepilite/ and format it.

_blinka/orange_pi_plus_2e.md
```diff
@@ -0,0 +1,58 @@
...
+blinka: true
+features:
+  - Wi-Fi
+  - Bluetooth/BLE
+  - Ethernet
+  - HDMI
+  - 40-pin GPIO
+---
+
+What’s Orange Pi PC？
```
makermelissa: Orange Pi PC again...

_blinka/orange_pi_lite.md
yeyeto2788: Ups, my bad. I have changed this as per your request.

_blinka/orange_pi_plus_2e.md
yeyeto2788: Same as before :p

yeyeto2788:
Of course, I'll fix it but I think the pushed images size for the large and small format are exactly those 😔 but I'll double check then.

I'll also  check the aspect ratio is correct as in the 2e.

Thanks for doing the review.

yeyeto2788:
Hey @makermelissa,

I have done the changes as you requested, sorry for the delay on it.

ladyada:
hi these commits seem to be malformed?
https://circuitpython.org/blinka

image

yeyeto2788:
Ups, my bad. Reference a `jpg` instead of a `png`. I have opened a PR fixing the issue.

Sorry for the inconveniences.

PR #391, opened 2020-01-27
Fix broken link for Orange Pi One

yeyeto2788:
As per https://github.com/adafruit/circuitpython-org/pull/383


makermelissa:
Thanks, but already fixed in #390.

PR #401, opened 2020-02-05
Simple Hero Management

jwcooper:
Moves moves the main hero images to their own folder and adds a _config.yml setting to set the hero image that is currently active.

Also adds 4 more hero images.


jwcooper:
Thanks!

PR #417, opened 2020-03-01
Update Absolute Newest and Past Releases download info

dhalbert:
- Add S3 download button for Past Releases, since future releases will only have assets on S3.
- Update download blurbs.


dhalbert:
image

PR #421, opened 2020-03-11
fix learn guide links for ItsyBitsy M0, Metro M4, and PyGamer

FoamyGuy:


FoamyGuy:
Sounds good to me. I'll make a new commit in a little while to put those back.

FoamyGuy:
I've added back the guide search links to the ones that I changed. I also found some other devices that have a good list of search results and added links for them as well.

_board/circuitplayground_bluefruit.md
```diff
@@ -50,7 +50,8 @@ on the go!
 
 ## Tutorials
 
-* [Circuit Playground Overview](https://learn.adafruit.com/adafruit-circuit-playground-bluefruit)
+* [Circuit Playground Bluefruit Overview](https://learn.adafruit.com/adafruit-circuit-playground-bluefruit)
+* [Projects and Guides](https://learn.adafruit.com/search?q=Circuit%20Playground%20Bluefruit)
```
tannewt: Instead of doing a search, there are pages that show every guide that is associated with a product: https://learn.adafruit.com/products/4333/guides

This is based on what products are in the sidebar and I'd expect it to give better results.

FoamyGuy: Neat! I had no idea these existed. I will change the search links later tonight.

tannewt: Yup! The only way I know how to get them is from the product page. Having them here will make them better known. Thanks for updating them!

makermelissa:
I noticed after merging in that the links are on the same line. Maybe we can make these bulleted if you'd like to submit another PR @FoamyGuy?

FoamyGuy:
@makermelissa Hmm I thought were in bulleted lists already. Do you know which ones you saw that were on the same line? I took a look at a few of them on circuitpython.org/downloads and the ones I saw do seem to be appearing in a bulleted list now. I can definitely make a new PR to fix any that need it though.

makermelissa:
Screen Shot 2020-03-12 at 6 44 50 PM

FoamyGuy:
Ah thank you. #422 fixes the CLUE links.

PR #423, opened 2020-03-16
Add note about UF2 bootloader being required for Arduino Nano 33 BLE

dhalbert:
https://forums.adafruit.com/viewtopic.php?f=60&t=158279&p=803678#p803678 requested a prominent note about this.


gigwegbe:
Hello @dhalbert,
I trying to flash Arduino ble 33 in order to install circuitpython on it. Can I use [serial converter cable](https://www.adafruit.com/product/70) for it? Thanks.

PR #445, opened 2020-04-04
Added RSS Icon + Added image to items

makermelissa:
I added an RSS icon that links to the new feed and updated the top menu spacing so it looked better with the new icon. I also added images to the description of the RSS items so that they should show up in the users' feeds and tested with NetNewsWire. Adding @jwcooper as a reviewer since it affects the styles.


makermelissa:
Thanks @kattni

PR #446, opened 2020-04-08
Add bootloader update section on board page

dhalbert:
Add a section on each board download page with links and information about updating the bootloader. There is a corresponding new data file, `_data/bootloaders.json`. Right now it needs to be updated manually, but should eventually be automated.

The section appears only for `atmel-samd` and `nrf` boards right now, but not for STM and other famiies for which there's a built-in bootloader or we don't have a good story yet.

Some third-party boards have not PR'd changes to the https://github.com/adafruit/uf2-samdx1 repo, or have done so only to the upstream repo. The bootloader section is omitted for boards for which we don't have links to bootloader updates.

Screenshots:

The story for `atmel-samd` boards is easy:
image

For nrf boards, the text is more cautionary:
image


makermelissa:
This looks great, however I'm concerned with removing the Gemfile.lock is that it may not compile on Github Pages after that.

dhalbert:
> This looks great, however I'm concerned with removing the Gemfile.lock is that it may not compile on Github Pages after that. I was able to generate the pages, but I believe it locks things to certain versions.

If I delete the `Gemfile.lock` file, it is re-created each time I ran `bundle exec jekyll serve` to check the build locally. So I thought this was just a lock file created by the local test webserver to prevent multiple webservers from running simultaneously, or something like that. I didn't know it had a bunch of versions in it, but it changed when I updated ruby stuff locally.

Info here: https://stackoverflow.com/questions/4151495/should-gemfile-lock-be-included-in-gitignore. So I should put it back?

dhalbert:
ok, I put `Gemfile.lock` back, after reading that stackoverflow link.

makermelissa:
The only other question I have is what are the adabot changes you included?

dhalbert:
I was going to say don't merge this yet, I need to test the bootloader some more. I just got a bug report.
I'm going to revert this for now.

makermelissa:
Ok, we can do another one and I'll hold off merging.

PR #478, opened 2020-05-16
contributing: Add weblate

jepler:
It looks roughly like this:

Screenshot_2020-05-16_17-54-51

There are other widgets we could choose instead:
https://hosted.weblate.org/widgets/circuitpython/

### DIFF
--- a/_layouts/contributing.html
+++ b/_layouts/contributing.html
@@ -23,6 +23,9 @@ layout: default
       Have an idea for a new driver or library?
       <a href="https://github.com/adafruit/circuitpython/issues">File an issue on the CircuitPython repo!</a>
     </p>
+    <p>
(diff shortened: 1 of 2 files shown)


dhalbert:
I like the bar graphs, one of these:
image, because it will be apparent to someone with a particular language skill what's needed.

I think that in combination with one of the "big" status badges conveys a lot of information.
image

jepler:
I don't like the bar graphs, as the name of the top language is cut off, and there's nothing obvious that we can do about it.

jepler:
I opened an issue with weblate about this: https://github.com/WeblateOrg/weblate/issues/3974

jepler:
Now using the bar graph style:

image

jepler:
@dhalbert can you rereview please?

PR #480, opened 2020-05-19
Add hiibot_bluefi board

BradChan:
Add hiibot_bluefi board definitions.


makermelissa:
Hi, could you limit the features to those in this guide?
https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-downloads? The list should only be limited to these 10 items and spelled exactly the same:

* Speaker
* Solder-Free Alligator Clip
* Feather-Compatible
* Battery Charging
* Display
* Wi-Fi
* Bluetooth/BTLE
* Robotics
* LoRa/Radio
* GPS

So for instance `BlueTooth 5 (nRF52840)` whould be changed to `Bluetooth/BTLE` and thing like `DC-DC` and `Optical sensor` aren't even on that list. Thanks. This helps keep our feature search to a reasonable length.

PR #481, opened 2020-05-21
Add Alorium Technology Evo M51

jpecor:


_board/aloriumtech_evo_m51.md
```diff
@@ -0,0 +1,49 @@
...
+board_id: "aloriumtech_evo_m51"
+title: "AloriumTech Evo M51 Download"
+name: "AloriumTech Evo M51"
+manufacturer: "Alorium Technology, LLC"
+board_url: "https://www.aloriumtech.com/evom51"
+board_image: "aloriumtech_evo_m51.jpg"
+date_added: 2020-5-21
+downloads_display: true
+blinka: false
+download_instructions: "BLINKA ONLY - url"
```
makermelissa, line 12: Can we remove this line? It looks like you used the template and for non-Blinka boards it should be removed. Thanks

jpecor: Yes.  I wasn't sure what to do with that line, so I left it in.  So, no problem removing it. 

Is that something I do?

makermelissa: Yeah, just remove it and push the change.

PR #483, opened 2020-05-24
Hide hiibot bluefi until it's added to CircuitPython

makermelissa:


BradChan:
The hiibot bluefi have added to CircuitPython since 5.4.0 Beta 0. Please show the hiibot bluefi board.
https://github.com/adafruit/circuitpython/releases.

PR #485, opened 2020-05-29
Added Odroid XU4 and XU4Q boards

makermelissa:


makermelissa:
Hmm, not sure if this one just kind of fell through the cracks.

PR #495, opened 2020-07-03
Use single space instead of 4 spaces

tannewt:
This makes the file less than 1MB and the existing GitHub API will work.


tannewt:
Done! Working on removing the filenames in the longer term by adding extension and language list.

dhalbert:
> Done! Working on removing the filenames in the longer term by adding extension and language list.

Yah, maybe sort of like how I do the bootloaders file; I have a template for the URL.

PR #505, opened 2020-07-17
Added message for chrome users that displays info about blocked UF2s

makermelissa:
Alert message will only show for chrome and includes a link to disabling protection in safe browsing.


ladyada:
let me know when merged i will test it out!

jepler:
This was fixed by google and can be backed out, right?

dhalbert:
I saw this as fixed even before this fix went on, so yes.

PR #508, opened 2020-07-24
Update Localization title and copy.

kattni:
@jwcooper Tagged you to verify that I didn't break any HTML in moving things around. 

@jepler Please check the content changes and see if you like them. Feedback is welcome and appreciated.


kattni:
Something is up with the checks, verified with Justin. Merging.

PR #515, opened 2020-08-08
Update kramdown version

dhalbert:
Address automated security vulnerability notice.


dhalbert:
@jwcooper it turns out this was a dependency mess. `kramdown` is actually listed twice in `Gemfile.lock`, and I missed one. Then, Trying to build locally to test failed, because `jekyll` 3.8.4 required an earlier version of `kramdown`. Updating to 3.9.0 implied a bunch of other updates as well. Also `kramdown` 2.x now splits out `kramdown-parser-gfm`, among other things, and that's a separate dependency.

This seemed like a nightmare to fix manually, but then I saw that if we used the `github-pages` gem, it would manage all these dependencies: it uses the new `kramdown`, uses `jekyll` 3.9.0, etc. I saw it mentioned in `Gemfile`, but it was commented out in favor of raw `jekyll`. I switched that around, deleted `Gemfile.lock`, did `bundle install`, rebuilt locally, and the site still looks fine.

Did that make sense? I wasn't sure why we weren't using that gem explicitly already.

jwcooper:
Yea, I think this all sounds good. Thanks for digging into it!

PR #517, opened 2020-08-09
Patch 1

tinkeringtech:
Modified tinkeringtech_scoutmakes_azul.md with updated details.


tinkeringtech:
Thank you! For future boards, do I creare and submit the .md file or will the CP.org admin create that page for me? I was slow in submitting to cp.org after adding to the adafruit CP core.

makermelissa:
It would definitely be helpful if you created it. Thanks.

PR #521, opened 2020-08-13
Per email communication from Seeed, there are no plans to support the…

jbfink:
… Wio Terminal's RTL8720DN in CircuitPython.

---

Wrote to Seeed about WiFi/BLE support in the Wio Terminal, received the following message:

 Hi
Thank you for like our wio terminal product.
We have a wiki to introduce the Circuitpython.   
But we have no plan to add Wifi and BLE on the RTL8720DN chip supported in
CircuitPython.  

Thanks for your time and patience.
Regards


dglaude:
This was to be expected, however I may have heard another story in a recent
Adafruit video. Maybe during a Show and Tell, Phil was relaying a question
from the chat about WiFi...

dhalbert:
A note I left for @ adr in Discord:
>If the RTL8720DN provides the HCI serial protocol to control it to use BLE, then it would be possible in the long run. I am implementing _bleio for HCI, targeting the ESP32, but another HCI-based BLE module should be easy to port to, with maybe essentially no changes. We would not do that work, but it would be possible.

tannewt:
I'll merge this change so that we don't promise anything. However, I do believe @ansonhe97 or someone else at Seeed is working on RTL8720DN Arduino support which should make it easier to add CircuitPython support. Our ESP32SPI coprocessor actually runs [Arduino based firmware](https://github.com/adafruit/nina-fw/) that someone may be able to port to the RTL8720DN Arduino.

PR #540, opened 2020-09-21
Squish whitespace to make the file smaller

tannewt:


jepler (approved):
Verified locally with `git show -b` that the only change was to whitespace.

PR #546, opened 2020-09-28
Added QT Py and more

makermelissa:
Other changes include:
 * unhiding TG-Watch-02A since there was enough information available online that it looks good.
 * I kept BD Micro M0 (which is not on their website anymore) since it's still a separate supported board in CircuitPython


tannewt:
The BDmicro Vina M0 was renamed the Vina SAMD21: https://github.com/adafruit/circuitpython/pull/3432

ladyada (approved):

PR #552, opened 2020-10-02
Added Metro ESP32-S2, STEMMA QT feature, cleaned up

makermelissa:
I added the following board:
 * adafruit_metro_esp32s2
 * unexpectedmaker_feathers2_prerelease

I couldn't find any differences between Feather S2 Prerelease and the regular Feather S2, so I hid the older one. We can always unhide if Unexpected Maker has a good reason for both to appear.

There are so many boards with STEMMA QT that I added it as a feature. I cleaned up other features where a 1-character difference made 2 appear like Feather-compatible and Feather-Compatible. I also checked the guide and template files and made sure everything matched.


ladyada (approved):

UnexpectedMaker:
Hi @makermelissa - The prerelease board has a different header pin layout. Different pins for I2C etc, and there are quite a few folks with them, so please make it visible again. Maybe down the line we can hide it.

makermelissa:
Sure thing @UnexpectedMaker. Do you have an image or info we can add to the board file?

UnexpectedMaker:
Hmm, all of that was already in CircuitPython.org.  Something has gone wrong .. seems the info for my pre-release and proper board have been merged. There is a listing only for the pre-release, but it's not using the pre-release image (green board).

Can you revert to my original commit? I had both boards setup with correct images and headings.

makermelissa:
Hmm, I didn't see an existing one for the prerelease. #555 will make the older one reappear as soon as it's approved and merged.

UnexpectedMaker:
Ok, once #555 (great commit number!) is merged, I'll pull and re-submit the board info again for both board revisions.
Thanks :)

PR #554, opened 2020-10-02
Updated pics and links for a few boards

makermelissa:
I updated the photos of a couple more boards with Adafruit versions, added another purchase link and replaced the OSH Badge photo with a better looking one.


ladyada (approved):

ladyada:
thank u :)

PR #556, opened 2020-10-02
FeatherS2 boards update

UnexpectedMaker:
Re-Added FeatherS2 pre-release board file and images and updated production board images.


makermelissa (changes requested):
Could you resize the images so they are a 13:10 width to height ratio so they match the other images on the site? 

Also, I think maybe you meant to add a space in "Unexpected MakerFeatherS2 Prerelease".

Otherwise this looks good.

UnexpectedMaker:
Woops, sorry, that's what you get when you rush!
I hope these are fine. The small ones were correct from what I can tell.

makermelissa (approved):
Looks good. Thanks.

PR #567, opened 2020-10-22
Added Adafruit store purchase links

UnexpectedMaker:
and feathers2.io web link so visitors can get more info
Also added purchase links to final board on the pre-release page.


makermelissa (approved):

makermelissa:
Thank you

PR #578, opened 2020-11-10
Add new board hiibot_iots2

BradChan:


makermelissa (changes requested):
Thanks for submitting this. Could you update the dimensions of your images so they blend in with the other images better. See https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/preparing-the-images for more details. Thanks.

PR #582, opened 2020-11-12
Show DynOSSAT-EDU boards

ecasadod:
We're ready to rock.

Added a Tindie link to purchase the kit and set the hardware repos to public.


makermelissa (changes requested):
Thanks for adding these. I was having trouble finding images/info. Would you mind shrinking down the large dynossat_edu_eps.jpg image? It's currently at 2948x2268. The rest of the images sizes are perfect.

ecasadod:
Of course, it seems I misplaced the "original" size in the "large" folder. Should be fixed now.

Thanks!

makermelissa (approved):
Excellent. Thank you!

PR #584, opened 2020-11-16
Remove non-standard features from list

makermelissa:
This was causing an additional 3 features to be added.


kattni (approved):

tannewt:
Want to add a Python script to check this in the CI? It should be easy to do and prevent extraneous features from getting checked in.

makermelissa:
That's a good idea. :)

PR #600, opened 2020-11-30
adding cp sapling markdown and image assets

skerr92:
Hey there! I am submitting the CP Sapling M0 for the CircuitPython.org site. I believe everything is in order. (There was some weird modes created for images with .iCloud extensions which I think I just need to click the download button in my computer file system if this is an issue for merging)

Let me know if I need to change anything else :)


skerr92:
I downloaded the images locally since iCloud cached them earlier. Re-committed and that made it so the only changes are the files I added for the CP Sapling. Should be good to go now.

makermelissa (approved):

PR #607, opened 2020-12-09
Update stm32f411ce_blackpill.md

Orel-A:
Changed incorrect product manufacturer and description. Added links for Arduino support and Docs.
Product image should be:
תמונה
Related URL: https://circuitpython.org/board/stm32f411ce_blackpill/


makermelissa (approved):
Thanks for updating. I think I added this originally because that's all the info I could find. :) Current product photo is an official one taken by Adafruit.

PR #611, opened 2020-12-28
manual release of 6.0.1

dhalbert:
Had to update version manually, due to change in `build_board_info.py` and structure of `_data/files.json`. Tried running new version in old tree, but ran into credential problems.
Also updated download counts, using December 2020 data.


tannewt (approved):
Looks good!

PR #616, opened 2021-01-11
Remember language selection and fix autoselection

Neradoc:
When a user selects a language in a download menu:
- the language selected is memorized in the browser
- all language menus on the page are synchronized (if possible)

When a page is loaded
- the saved language is recovered and used to set the download menus
- if no language was saved it pulls the favorite languages from the browser as before

It's implemented with localStorage which seems more fitting to me for a static site, but could use cookies.

This also includes a fix to the selection of the menus' language. It now iterates on user languages in their order of preference to pick the best match.
Also, when the browser gives "`<language>-<region>`",  it first tries a full match, then tries to match `<language>` to the locales available on the site. And I tossed a couple of toLowerCase() in there for the string comparison.
- the browser's "fr-FR" will match the site's "fr"
- the browser's "id" will match the site's "ID" (Indonesian)

What it doesn't do is reduce the site's locale string to it's `<language>` for comparison, because I didn't want "en-UK" to accidentally match "en-x-pirate". Making that work would require to make sure that the default variant is first in the language lists, or otherwise identified.
- the browser's "en-UK" will not match the site's "en-US"
- the browser's "de-AT" will not match the site's "de-DE"

### DIFF
--- a/assets/javascript/download.js
+++ b/assets/javascript/download.js
@@ -4,60 +4,77 @@ document.addEventListener('DOMContentLoaded',function() {
     select.onchange = languageSelectHandler;
   });
 
-  var languages = null;
+  // get the language from memory
+  var storedLanguage = null;
   if (window.localStorage) {
-    var storedLanguage = localStorage.getItem("language");
+    storedLanguage = localStorage.getItem("language");
     if (storedLanguage != null && storedLanguage != "") {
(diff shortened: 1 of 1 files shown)


jwcooper (approved):
This looks great, thank you!

PR #629, opened 2021-01-25
Revert github pages dependency changes

makermelissa:
Ok, trying to decrypt the error message, I think what it is saying is that the dependency versions under github pages don't match, so I likely modified this in the wrong place, so I'm reverting that one change I did manually.


TheKitty (approved):

makermelissa:
Yay! This seems to have resolved the dependency issue!

TheKitty:
Yay! I'll wait a bit and check the website again

PR #632, opened 2021-01-31
remove 5.x bundle boilerplate

dhalbert:
The text about the bundle was there, but there was no link.

This might be automated in the future, but fixing quickly for now.


tannewt (approved):
Thanks!

PR #640, opened 2021-02-19
Add Silicognition M4-Shim board

xorbit:
Let me know if I missed anything.  This is a placeholder that will be updated later with release pictures and purchase links.


makermelissa (changes requested):
Hi, images look great. However, could you please keep the features to those in the list on this page: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-downloads

xorbit:
Hi Melissa,

Sorry about that, I was wondering what was allowed for features but I hadn't seen that page.  Thanks for the link.
I was also wondering about how strict "Feather-Compatible" is.  It has an unusual board outline and doesn't support battery charging, but it's definitely intended to be used with the Feather ecosystem.

I also changed the copy to have a "Learn More" section since the board is not for sale yet.

makermelissa:
Thanks. For feather-compatible, I usually go by "will it work with featherwings?

makermelissa (approved):

PR #642, opened 2021-02-24
Added pages for Pimoroni's RP2040 boards

ZodiusInfuser:
This covers the Tiny 2040, Keybow 2040, and PicoSystem.

I have tested these locally as instructed, but did not see a download .uf2 option as on Pico etc, nor a flag to make that appear.

Also, the instructions for what tags to include mentioned a list of ones to use, but did not actually give a list. Could you point me to the list so I can update the tags accordingly.


makermelissa (changes requested):
Thanks for adding. Could you adjust the Features list so that it matches only the items in the list available at this page: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-downloads#header-3034486-5.

ZodiusInfuser:
Thanks. I have updated these now.

May I suggest including a link to that guide in the template.md. It was not obvious to me that the comment that said "this list" was referring to the options below, in part because I was expecting there to be more choices that would align with our products.

tannewt (approved):
Looks good to me! Thank you!

makermelissa (approved):

PR #654, opened 2021-03-19
Added Unexpected Maker TinyS2 board

UnexpectedMaker:
Added board info and images for my new TinyS2 ESP32-S2 board


makermelissa (changes requested):
Hi, thanks for adding this. Could you please remove TinyPICO-Compatible from the features? Here's a guide that has the values you can use: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-downloads.

We need to limit these because when we didn't the available options in the filters on circuitpython.org got out of control and wasn't very useful anymore. Thanks.

UnexpectedMaker:
> Hi, thanks for adding this. Could you please remove TinyPICO-Compatible from the features? Here's a guide that has the values you can use: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-downloads.
> 
> We need to limit these because when we didn't the available options in the filters on circuitpython.org got out of control and wasn't very useful anymore. Thanks.

Oh yeah, sure! I actually didn't mean to leave that in the features list after I added it to the specs below. Sorry :(

makermelissa (approved):
No worries. Thank you!

PR #655, opened 2021-03-24
fix files.json trailing commas

Neradoc:
Fixes error on `json.load()`.

### DIFF
--- a/_data/files.json
+++ b/_data/files.json
@@ -67,7 +67,7 @@
      "en_US",
      "ja",
      "pl",
-     "de_DE",
+     "de_DE"
     ],
     "modules": [
      "analogio",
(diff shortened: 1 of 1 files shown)


makermelissa (commented):
Thanks for fixing this. Since this file is auto-generated, is this likely to be an issue in the future?

jepler (approved):
Spot checked and the changes look okay

dhalbert:
> Thanks for fixing this. Since this file is auto-generated, is this likely to be an issue in the future?

This was my fault. It doesn't normally contain trailing commas. I had to edit it by hand due to https://github.com/adafruit/circuitpython/issues/4470

dhalbert (commented):
:+1:

PR #656, opened 2021-03-24
Sort the language menus by language name

Neradoc:
The language menus are in an arbitrary order, making it potentially confusing when looking for one. The order used is the one from the files.json file, which in turn depends on the way languages are listed with `os.scandir` in the circuitpython repository (`build_board_info.py`).

The displayed language names are in the `board.html` template. To sort by name the code uses Liquid's {%capture%}  and split to make a new list with the name and the language code, so it can be sorted. This requires a little bit of manipulations to avoid whitespaces interfering with the sort.


tannewt (approved):
Awesome! Thank you!

PR #659, opened 2021-03-30
Update kramdown to 2.3.1

dhalbert:
Address Dependabot security alert. I am flying a little blind here, but we're just upgrading from 2.3.0 to 2.3.1.


jepler (approved):
The "s" in "internet" is for "security".

Issue #587, opened 2020-11-17
Clicking a filter doesn't update the url


tannewt:
Click a feature like "Wifi". The board list updates but the url doesn't, so it can't be shared.

makermelissa:
Fixes via #670.

PR #670, opened 2021-04-13
update the URL with feature filters

Neradoc:
Fixes issue #587


makermelissa (approved):
I didn't test, but this looks change looks like a good one.

makermelissa:
Tested on the live site and it works great. Thank you!

PR #671, opened 2021-04-13
remove the 6.2.0-rc.0 release

Neradoc:
Manual fix to remove the unstable release. The `files.json` was generated by the fixed script running on 6.2.0, and copying the downloads stats from the current one. Also for some reason `rp2pio` was missing and now isn't. Ok then.


jwcooper (approved):
Approving - can't review diff inline in GitHub, can always roll it back. File looks good.

PR #677, opened 2021-04-15
Github Actions - switch to ruby/setup-ruby and update dependencies

jwcooper:
Fixes: https://github.com/adafruit/circuitpython-org/issues/665


dhalbert (approved):
I tested this locally and it updated my dependencies and produced a working local website. Thanks!

PR #679, opened 2021-04-16
Lazy load images if browsers support the loading attribute

jwcooper:
Speeds up downloads and blinka page loading times on slower connections. Older browsers will just ignore the new attribute.

Browsers supported: https://caniuse.com/loading-lazy-attr


ladyada (approved):

dhalbert (approved):
great!

PR #680, opened 2021-04-17
Added ARAMCON2 Badge

bmeisels:
See https://github.com/adafruit/circuitpython/pull/4627


makermelissa (changes requested):
Thank for submitting. I see a couple of small things. For download instructions, either just provide a blank value. The images appear to be a bit wide. We have a guide on the sizing here: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/preparing-the-images. They should be a ratio of 13:10 for width to height.

makermelissa (approved):
Looks great. Thanks for making the changes.

PR #683, opened 2021-04-26
Fix typos

2231puppy:
Hi, I just noticed a few grammatical things on the index page. Nothing big, just making it easier on the ol' brain ;)


kattni (approved):
Thank you for the updates! It definitely reads better this way.

PR #686, opened 2021-05-06
Updated the link to the correct place to order this version.

eddieespinal:
I'm currently crowdfunding this version on GroupGets.com. The link was taking people to the older version which doesn't support circuitpython.


makermelissa (approved):
Thanks, looks good.

PR #690, opened 2021-05-13
Added pages for Pimoroni's PicoLipo boards 

ZodiusInfuser:
This covers the 16mb and 4mb versions of the PicoLipo board. The former should be launching this week, so if we would get this merged soon, that would be greatly appreciated!


makermelissa (changes requested):
Thanks for submitting. Could you please resize the photos according to this guide: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/preparing-the-images? Otherwise it looks good.

ZodiusInfuser:
Hi. Could you elaborate? We've prepared pictures exactly as described on the github page, and ran them through the recommended webpage to compress them.

`Provide 3 images. An original high-quality image. A smaller image (300 px width), and a larger image (700 px width) in each respective directory (assets/images/boards/{small large original}) and process them in something like https://squoosh.app/ to reduce file size. If you only have one image, place it in the 'original' folder.`

This is the exact same process we did for our previous boards without issue.

makermelissa (approved):
Oh yeah, fair enough. I just checked the other images (same dimensions) and they look fine. Maybe the image sizing is a bit more resilient than I thought. :)

ZodiusInfuser:
Thanks @makermelissa! I'm curious now what the issue was that made you flag this. That guide does mention different dimensions to what this repo says, so perhaps ones needs to be updated?

makermelissa:
Oh, it's recommended to have the images at a 13:10 ratio (like the other images on the site), but yours are 4:3. It looks fine either way.

PR #691, opened 2021-05-14
Update BHB description

ajs256:
It seems that there was a copy-paste from another board that was not updated.


tannewt (approved):
Thank you!

PR #703, opened 2021-06-01
Added NanoESP32 S2 w/WROVER & reword WROOM version

reppad:
Here is the download page for the new CircuitPython build for the Muse lab NanoESP32-S2 w/WROVER.
This pull request also update the download page of the original NanoESP32-S2 that is now named NanoESP32-S2 w/WROOM.
For reference, the related PR in CircuitPython repo: https://github.com/adafruit/circuitpython/pull/4834

### DIFF
--- a/_board/muselab_nanoesp32_s2_wroom.md
+++ b/_board/muselab_nanoesp32_s2_wroom.md
@@ -1,0 +1,25 @@
+---
+layout: download
+board_id: "muselab_nanoesp32_s2_wroom"
+title: "NanoESP32 S2 w/WROOM Download"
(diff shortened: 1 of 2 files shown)


makermelissa:
We'll need to update the _data/files.json to update the board names (https://github.com/adafruit/circuitpython-org/blob/master/_data/files.json#L14864-L14944) or we will have an Unknown Board appear for the old name. I see they are already updated in circuitpython itself, so that shouldn't really be an issue.

makermelissa (approved):
Ok, approving because this looks good. I'll submit a follow-up PR to fix the resulting unknown board.

PR #705, opened 2021-06-03
remove the 6.3.0-rc.0 release

Neradoc:
Since https://github.com/adafruit/circuitpython/pull/4594 is not part of 6.3.0, this has to be done manually again.

Created by a script that filters at the json level, and checked by comparing with `git diff --histogram` which gives much more relevant results than the default algorithm: it correctly identifies the missing entries in the versions array rather than the mess of comparison of unrelated parts that the default algorithm does.

https://gist.github.com/Neradoc/8b5819a9aebf85f6f9824037b184968b


makermelissa:
Ok, it sounds like this removes 6.3.0-rc.0 from under Latest unstable release. Does it replace it with anything else or just remove latest unstable release altogether?

dhalbert:
There is no 7.0.0-alpha release, but I am preparing one and it should be out soon (today or a few days).

makermelissa (approved):
Ok cool. Just tested locally and it looks good.

PR #712, opened 2021-06-07
Rename master to main

jwcooper:
This depends on solving all outstanding issues in the following issue:
https://github.com/adafruit/circuitpython-org/issues/711


jwcooper:
> There is a reference in `images.yml` as well
> 
> https://github.com/adafruit/circuitpython-org/blob/df9dea3835211265f6104372c3b8c425cec675df/.github/workflows/images.yml#L8

I've updated that to use v2 of the checkout action.

lesamouraipourpre:
> 
> 
> > There is a reference in `images.yml` as well
> > https://github.com/adafruit/circuitpython-org/blob/df9dea3835211265f6104372c3b8c425cec675df/.github/workflows/images.yml#L8
> 
> I've updated that to use v2 of the checkout action.

Shouldn't this be
` - uses: actions/checkout@main` for the change to main

dhalbert:
`- uses: actions/checkout@<some-tag>` is a tag or branch name from https://github.com/actions/checkout, not related to our `main`. They have switched to `main` in that repo, but usually you want to stick to a particular major version for a particular GitHub action.

lesamouraipourpre:
> 
> 
> `- uses: actions/checkout@<some-tag>` is a tag or branch name from https://github.com/actions/checkout, not related to our `main`. They have switched to `main` in that repo, but usually you want to stick to a particular major version (e.g., `v2`) for a particular GitHub action.

Slight misunderstanding on my part. As Dan said, I think `- uses: actions/checkout@v2` is the current recommended usage, even though the last release was in November.

jepler:
What's needed to move this forward?

jwcooper:
As of the PR release, I think we just needed to check that any bots pushing to this repo were using main instead of master? I'm not really sure. The changes in this PR should still be OK, but we'll need to update any newly added board files still referencing master.

jwcooper:
I've just synced it up with master and updated any new files. This PR is now back up to date.

@makermelissa @tannewt is there anything else we need to do to get this one in? Any adabot changes?

FoamyGuy:
On the Blinka devices there is a link near the bottom of their MD files:
image

The URL for that link appears to have the `master` branch coded into it. So that would need to get updated to main in each md file I think. i.e. 

https://github.com/adafruit/circuitpython-org/blob/30484e5690bcbccc76c2c733c68e4ece135748f0/_blinka/beaglebone_black.md?plain=1#L54

There are a few places in this guide where the same thing could get updated: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-blinka

Maybe some or all of the non-blinka microcontrollers have this as well I haven't done a thorough check, but Feather Express M4 does have a similar link that could get updated.

This process could probably be automated fairly easily instead of trying to do all of them "by hand"

jwcooper:
@FoamyGuy those should all be synced up now as well.

FoamyGuy:
Ah, that is right, I see them in the changes now. @jwcooper thank you.

Updating the accompanying example file references on these pages: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-downloads

https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-blinka

were the only two places I found in Learn Guides that may need updated to match when the time comes.

jepler:
still another conflict :-/

makermelissa:
I changed the name to raspberry_pi_cm4io to reflect the fact that we have multiple carrier boards that can run the compute module 4.

jwcooper:
@jepler @makermelissa this should be ready again.

makermelissa (approved):
Looks good to me. Jeff is out on vacation.

dhalbert:
I don't think we need to wait for Jeff. It's been reviewed. You could merge and we will see that it works or not, and fix it.

kattni:
I'm ok with merging this.

dhalbert:
I renamed the default branch now to `main`.

PR #715, opened 2021-06-09
Add Kaluga v1.3

jepler:


makermelissa (changes requested):
Could you update the date on the 1.3 board to today? Otherwise it looks good.

makermelissa (approved):

PR #729, opened 2021-07-13
Add USB-C as a feature filter

dgriswo:
USB-C is a pretty notable feature to search on.

I also considered a feature for USB Micro, but then I felt it would be excluding USB A, USB Mini, Castellated, and pin headers.  I felt adding each possible USB connectivity would make the filter list lengthier and a little awkward.

I also corrected the markdown syntax on a board to remove the "Wi-Fi Bluetooth/BTLE" category. It now filters into each category correctly.

### DIFF
--- a/_board/TG-Watch.md
+++ b/_board/TG-Watch.md
@@ -11,6 +11,7 @@ features:
   - Display
   - Bluetooth/BTLE
   - Battery Charging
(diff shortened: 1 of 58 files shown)


makermelissa (approved):
Makes sense to me since there are so many of them.

PR #732, opened 2021-07-30
Add Breadboard-Friendly and Arduino Shield Compatible.

dgriswo:
Added Breadboard-Friendly and Arduino Shield Compatible features.

Breadboard-Friendly is defined as parallel pins with at least one socket open on each side of the board (Assuming single, standard breadboard) and minimal non-critical perpendicular pins.  (i.e. 3V3, Vcc, Vusb, GND, En, Reset not perpendicular)

As I worked through the boards, I found some boards definitions missing features and corrected those.


makermelissa (approved):
Thanks, this should be helpful.

PR #750, opened 2021-09-05
Lolin S2 Mini board download page updates, features & TinyUF2 flashing instructions

durapensa:
Probably this PR should remain on hold until `tinyuf2-lolin_s2_mini-...zip` with `combined.bin` is available at https://github.com/adafruit/tinyuf2/releases ([PR#142](https://github.com/adafruit/tinyuf2/pull/142), but latest release was on 2021-07-30). For review in any case.

### DIFF
--- a/_board/lolin_s2_mini.md
+++ b/_board/lolin_s2_mini.md
@@ -13,15 +13,42 @@ features:
 
 ### Features
 
-- based ESP32-S2FN4R2 WIFI IC
-- Type-C USB
+- ESP32-S2FN4R2 WiFi SoC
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thanks

durapensa:
@Neradoc pointed out on Discord that a UF2 bootloader download section exists conditionally in [board.html](/adafruit/circuitpython-org/blob/master/_includes/download/board.html) based on data in [bootloaders.json
](/adafruit/circuitpython-org/blob/master/_data/bootloaders.json)

I'd need some guidance to update both, particularly considering that ESP32-S2 boards would be new to both.

PR #753, opened 2021-09-08
Add "Install/Repair UF2 Bootloader" section for ESP32-S2 boards

### LINKED ISSUE #752: New "Install UF2 Bootloader" section in download/board.html?
[_includes/download/board.html](/adafruit/circuitpython-org/blob/master/_includes/download/board.html) contains a section `<div class="section unrecommended"><h3>UF2 Bootloader</h3>...` with conditionals  `{% if bootloader_board.family == 'nrf52840' %}` and `{% if bootloader_board.family == 'atmel-samd' %}` that provides *bootloader updating* instructions.

For any board with with TinyUF2 in flash, a section to *install bootloader* would be helpful, particularly for those boards that do not have TinyUF2 factory-flashed. "Install UF2 Bootloader" sections exist in Adafruit Learning System articles for [FunHouse](https://learn.adafruit.com/adafruit-funhouse/install-uf2-bootloader), [MagTag](https://learn.adafruit.com/adafruit-magtag/install-uf2-bootloader) and [Metro ESP32-S2](https://learn.adafruit.com/adafruit-metro-esp32-s2/install-uf2-bootloader), for the case of bootloader corruption or overwrite (e.g. if the board is esptool-flashed with Arduino firmware and the user wishes to return to using CircuitPython).

For the [Lolin S2 Mini](https://circuitpython.org/board/lolin_s2_mini/), I wrote a "Flashing UF2 Bootloader" section in  [_board/lolin_s2_mini.md](/adafruit/circuitpython-org/blob/master/_board/lolin_s2_mini.md), but would a new section in [_includes/download/board.html](/adafruit/circuitpython-org/blob/master/_includes/download/board.html) better handle cases like this one?

durapensa:
Fixes #752

Checked for typos, tested with `bundle exec jekyll serve`, and verified that download links are correctly generated on the Lolin S2 Mini page. Have not manually tested the other affected 21 pages for ESP32-S2 boards, but mappings in bootloaders.json should be correct.

### DIFF
--- a/_includes/download/board.html
+++ b/_includes/download/board.html
@@ -168,11 +168,11 @@ By the way, boolean operation precedence is right to left! (yeesh)
     <strong>The bootloader allows you to load CircuitPython, Makecode, and Arduino programs.
       The bootloader is not CircuitPython.</strong>
   </p>
+
+{% if bootloader_board.family == 'esp32s2' %}
   <p>
     It is not necessary to reinstall the bootloader unless a BOOT drive is not visible when in UF2 bootloader mode (FTHRS2BOOT, MAGTAGBOOT, HOUSEBOOT, etc.), in which case the UF2 bootloader may need to be repaired.
   </p>
-
-{% if bootloader_board.family == 'esp32s2' %}
   <p>
     If a UF2 bootloader has never been installed on the board, or the UF2 bootloader was removed by erasing or overwriting the flash, the UF2 bootloader should be installed in order to flash .uf2 files onto the board.
   </p>
--- a/_includes/download/board.html
+++ b/_includes/download/board.html
@@ -160,7 +160,61 @@ By the way, boolean operation precedence is right to left! (yeesh)
 {% assign bootloader_id = bootloader_board.bootloader_id %}
 {% if bootloader_version and bootloader_id and bootloader_board %}
 <div class="section unrecommended">
-  <h3>UF2 Bootloader</h3>
+  <h3>Install/Repair UF2 Bootloader</h3>
+  <p>
+    Latest version: {{ bootloader_version }}
+  </p>
+  <p>
+    <strong>The bootloader allows you to load CircuitPython, Makecode, and Arduino programs.
+      The bootloader is not CircuitPython.</strong>
+  </p>
+  <p>
+    It is not necessary to reinstall the bootloader unless a BOOT drive is not visible when in UF2 bootloader mode (FTHRS2BOOT, MAGTAGBOOT, HOUSEBOOT, etc.), in which case the UF2 bootloader may need to be repaired.
+  </p>
+  <p>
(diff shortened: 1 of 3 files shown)


_includes/download/board.html
```diff
@@ -160,7 +160,61 @@ <h3>Past Releases</h3>
...
+  <p>
+    <strong>The bootloader allows you to load CircuitPython, Makecode, and Arduino programs.
+      The bootloader is not CircuitPython.</strong>
+  </p>
+  <p>
+    It is not necessary to reinstall the bootloader unless a BOOT drive is not visible when in UF2 bootloader mode (FTHRS2BOOT, MAGTAGBOOT, HOUSEBOOT, etc.), in which case the UF2 bootloader may need to be repaired.
+  </p>
+  <p>
+    If a UF2 bootloader has never been installed on the board, or the UF2 bootloader was overwritten by erasing or overwriting the flash, the UF2 bootloader should be installed in order to flash .uf2 files onto the board.
+  </p>
```
dhalbert, line 176: Could you move this inside the `esp32s2` section? On non-ESP32S2 boards, there is always a UF2 bootloader (except in very special circumstances), so I think this is a little confusing for other board families

dhalbert (changes requested):
Thank you for doing this!

It would be good if someone else who is quite familiar with ESP32-S2 reviewed this as well. I never use the UF2 bootloader, and haven't used the WebSerial ESPTool.

makermelissa (commented):
Overall this looks good. I think after adding the changes Dan suggested, it should be good.

_includes/download/board.html
durapensa: Fixed in e4474dfa5a56a46ffff976f758d955d20a786572

_includes/download/board.html
```diff
@@ -160,7 +160,62 @@ <h3>Past Releases</h3>
...
+  <p>
+    Latest version: {{ bootloader_version }}
+  </p>
+  <p>
+    <strong>The bootloader allows you to load CircuitPython, Makecode, and Arduino programs.
+      The bootloader is not CircuitPython.</strong>
+  </p>
+  <p>
+    It is not necessary to reinstall the bootloader unless a BOOT drive is not visible when in UF2 bootloader mode (FTHRS2BOOT, MAGTAGBOOT, HOUSEBOOT, etc.), in which case the UF2 bootloader may need to be repaired.
+  </p>
```
dhalbert, line 173: Move this paragraph down too into the ESP32-S2 section, because if the bootloader doesn't show on another board, it's much harder to re-flash it (you may need a J-Link).

Basically, on the ESP32-S2, the bootloader is a "extra" thing, but on most other boards, it is permanently burned on the chip and is essential to its normal operation. So if it didn't appear, there is something really seriously wrong.

dhalbert (changes requested):
Thanks, just one more adjustment!

dhalbert (approved):
Thanks for your patience!

PR #754, opened 2021-09-09
Lolin S2 Mini - added more specs useful for makers

durapensa:
Low-priority.

I'm happy to templatize this and other MCUs, modules, other components, etc., for consistency on board pages, and for anything that might be useful in the "Downloads" search field.

For search perhaps by using the same mechanism as, e.g., `data-id="feather_nrf52840_express"` ... `data-features="Feather-Compatible,Battery Charging,Bluetooth/BTLE,Breadboard-Friendly"` and adding a `data-mcu` or `data-components` field to include things like `nrf52840` or `ESP32-S2FN4R2`, and/or other components, particularly useful for searching when MCU/component names do not appear in `data-name` nor elsewhere.

### DIFF
--- a/_board/lolin_s2_mini.md
+++ b/_board/lolin_s2_mini.md
@@ -16,9 +16,19 @@ features:
 
 - ESP32-S2FN4R2 WiFi SoC
     - Xtensa® single-core 32-bit LX7 microprocessor, up to 240 MHz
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Looks good. Thanks!

PR #758, opened 2021-09-13
Note error on Lolin S2 Mini v1.0 board silkscreen

evildave666:
Putting a note on the S2 Mini page about an error on the silkscreen markings on v1.0 of the board.
GPIO12 and 13 are reversed from what they should be.


ladyada (approved):
sure, who knows!

PR #766, opened 2021-09-21
migrate family & bootloader_id to board.md files, add mcu family search"

durapensa:
A cleaned-up version of #756 (supersedes that PR), including discussed migration of `family` & `bootloader_id` to `board.md` files per @makermelissa's suggestion. 

`bootloaders.json` now contains only bootloader port (family) & version information, and no longer contains board information.

Tested locally such that `_includes/download/board.html` points to the correct downloads for CircuitPython & UF2 bootloader, and search/filter works as in #756.


makermelissa (changes requested):
The MCU family is showing `undefined`. I think the fix may be to add an if statement to the code logic, but I'm not entirely sure.
Screen Shot 2021-09-21 at 11 26 04 AM

durapensa:
I've been trying to track that down, and it should be resolved now (fix in next push).

makermelissa (changes requested):
Thanks, the `undefined` option is gone for me now. However, it looks like the filters on the Blinka page have stopped functioning.

durapensa:
I didn't test that! ;-) Nice catch.

Fix was: rather than add conditionals to `assets/javascript/downloads.js`, just make it happy by supplying it with empty `mcufamily` data from `blinka.html`.

makermelissa (approved):
Thanks for the quick response. It appears to be working great!

PR #770, opened 2021-09-26
Fix community bundle link

Neradoc:
The link for the Community Bundle pointed to the Circuitpython Organisation Bundle (which should probably get a link too some day).


FoamyGuy (approved):
looks good to me. Tested locally. Thanks for the fix!

PR #772, opened 2021-09-29
feat: add Blues Wireless Swan

bsatrom:
Related to an official board support PR [here](https://github.com/adafruit/circuitpython/pull/5403).


makermelissa (approved):
Thanks, looks good.

PR #773, opened 2021-10-04
Renamed Nordic to Nordic Semiconductor for 2 boards

jpconstantineau:
Renamed the "Manufacturer" for 2 boards to match the actual name of the manufacturer.
Both the PCA10059 and PCA10100 had only "Nordic" as manufacturer while the PCA10056 had "Nordic Semiconductor"

This caused the 3 boards to show up as 2 different manufacturers in the list of filters, which obviously is redundant and incorrect.


makermelissa (approved):
Thanks. Looks good to me.

PR #784, opened 2021-11-02
New Library links for Community bundle

jwcooper:


libraries/index.html
```diff
@@ -104,11 +104,28 @@ <h3>Examples Bundle</h3>
     <section>
       <h2>The Community Bundle</h2>
       <p>
-      The libraries in the bundles above are officially supported.
-      Additional libraries written and supported by community members are available in the
-      <a href="https://github.com/adafruit/CircuitPython_Community_Bundle">Community Bundle</a>.
+        The libraries in the bundles above are officially supported.
+        Additional libraries written and supported by community members are available in the
+        <a href="https://github.com/adafruit/CircuitPython_Community_Bundle">Community Bundle</a>.
```
kattni, line 109: I think we should update the copy a bit as well, and point folks to the releases for getting the 6.x bundle we opted not to include in the links below. Consider something like what I suggested here.

```suggestion
        The libraries in the bundles above are officially supported by Adafruit.
        Additional libraries written and supported by community members are available in the
        <a href="https://github.com/adafruit/CircuitPython_Community_Bundle/">Community Bundle</a>.
        </p>
        <p>If you are looking for the 6.x Community Bundle, you can find it on 
        <a href="https://github.com/adafruit/CircuitPython_Community_Bundle/releases/">GitHub</a>.
```

kattni (changes requested):
Thanks so much for doing this! I really think making it easier to find the Community Bundle is an important thing. I really appreciate it.

I suggested a change to the copy below. I _think_ it's formatted properly, but please verify before accepting the suggestion.

kattni (approved):
Thanks so much for adding this!

PR #787, opened 2021-11-16
Added info for missing boards + updates

makermelissa:
This should fix all those pesky unknown boards.


_board/ai_thinker_esp32-c3s.md
```diff
@@ -0,0 +1,45 @@
...
+title: "ESP-C3-32S Download"
+name: "ESP-C3-32S"
+manufacturer: "Ai-Thinker"
+board_url: "https://docs.ai-thinker.com/en/esp32c3"
+board_image: "ai_thinker_esp32-c3s.jpg"
+date_added: 2021-11-15
+family: esp32c3
+features:
+  - Breadboard-Friendly
+  - Wi-Fi
```
dhalbert, line 13: Can you add BLE to this?

I note the feature list says `Bluetooth/BTLE`. `BTLE` is rarely used as an abbrevation; `BLE` is much more common. Perhaps we should fix other boards that use `BTLE`.

dhalbert (changes requested):
Thanks for doing all these! Just one addition.

_board/ai_thinker_esp32-c3s.md
makermelissa: Added, changing it to `BLE` is fine, but should probably be its own PR.

dhalbert (approved):
:+1:

PR #790, opened 2021-11-16
support ru (Russian) downloads

dhalbert:
We are now building boards with a Russian translation. There is no release yet, but this will get set for it.


makermelissa (approved):
Thanks

PR #794, opened 2021-11-30
Update to Ruby 3 and add RPI CP boards

tannewt:
Run the Jekyll build on pull_request as well to make sure it works


makermelissa (dismissed):
Looks good. Thanks for updating the URLs.

makermelissa:
FYI, looks like the build action is failing.

jepler:
deploy should be guarded by an if: condition, then?

tannewt:
I think the build step is doing fancier things than just building the site like fetching and committing new files. Could we split the data update from the build so that 1) we don't build on a schedule and 2) everyone can build on their branch.

tannewt:
> I think the build step is doing fancier things than just building the site like fetching and committing new files. Could we split the data update from the build so that 1) we don't build on a schedule and 2) everyone can build on their branch.

I'm wrong about this. The build step was fine. I'll add a condition on the deploy.

tannewt:
Build is green @makermelissa. I had to prevent the deploy when running on a PR.

makermelissa (approved):
Awesome, that works for me.

makermelissa:
Not sure if you're waiting on Justin to also review.

jwcooper (approved):

PR #796, opened 2021-11-30
Add overview page that enumerates ways to connect

tannewt:
This includes TestFlight links for the apps.

### DIFF
--- a/connect/index.md
+++ b/connect/index.md
@@ -1,0 +1,44 @@
+---
+layout: default
+---
+
+# Overview
+
+Starting with CircuitPython 7, there are a couple of ways of connecting to a CircuitPython board and editing files. We call these "workflows". The classic workflow is over USB where the CircuitPython board appears as a `CIRCUITPY` drive with all of the code files on it. The second way to connect is wirelessly over Bluetooth Low Energy (BLE). Unlike USB, BLE works better with phones and tablets.
+
+## USB
+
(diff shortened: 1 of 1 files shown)


tannewt:
I'm not sure where we want this to live in the menu. @tylerdcooper may have ideas on where to add a "Connect" link.

tannewt:
We can merge this before adding a link if we want to see the existing page at https://circuitpython.org/connect/

makermelissa (approved):
Looks good to me.

tylerdcooper:
@tannewt Any reason why this shouldn't just be a page on the learn guide in getting started? Our biggest issue right now is that the top nav is getting quite full.

tannewt:
@tylerdcooper I think it should be in the learn guide as well.

However, I think we need lightweight pages for linking out to things on circuitpython.org. Learn is prose heavy, harder to link to and harder for others to contribute to.

Maybe we should change the get started link to go to a page that links out to the connect page, tools page and starting learn guides.

tylerdcooper:
@tannewt I agree, I think our best bet is to switch the getting started link to a page on circuitpython.org that then links out to pages like this and the learn guide. I'll work on a mockup and share here. Maybe you could start putting together a page including all of the things you think we should include on the getting started page to help me make sure I don't miss anything?

tannewt:
I think @kattni would have good ideas about what to put on a "getting started" launch pad page too. I certainly think the "Getting started with circuitpython" and "circuitpython essentials" guides should be linked.

PR #797, opened 2021-12-01
Common layout and style connect page

jwcooper:
This PR adds a common layout for content type pages. The /connect page is now styled to match the other content pages as well.

### DIFF
--- a/_layouts/common.html
+++ b/_layouts/common.html
@@ -2,7 +2,7 @@
 layout: default
 ---
 
-<div class="common-layout">
+<div id="common-page" class="common-layout">
   <div class="content">
     {{ content }}
   </div>
--- a/_layouts/common.html
+++ b/_layouts/common.html
@@ -1,0 +1,9 @@
+---
+layout: default
+---
+
+<div id="libraries-page" class="common-layout">
+  <div class="content">
+    {{ content }}
+  </div>
+</div>
--- a/_layouts/common.html
+++ b/_layouts/common.html
@@ -2,7 +2,7 @@
 layout: default
 ---
 
-<div id="libraries-page" class="common-layout">
+<div class="common-layout">
   <div class="content">
     {{ content }}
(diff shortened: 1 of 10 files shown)


_layouts/common.html
```diff
@@ -0,0 +1,9 @@
+---
+layout: default
+---
+
+<div id="libraries-page" class="common-layout">
```
makermelissa, line 5: Did you mean to leave the id=libraries-page" for all pages?

_layouts/contributing.html
```diff
@@ -1,59 +1,57 @@
...
-      <a href="https://learn.adafruit.com/contribute-to-circuitpython-with-git-and-github">Contributing to CircuitPython with Git and Github</a>.
-      You can also find us in the #circuitpython channel on the
-      <a href="https://adafru.it/discord">Adafruit Discord</a>.
-    </p>
-    <p>
-      Have an idea for a new driver or library?
-      <a href="https://github.com/adafruit/circuitpython/issues">File an issue on the CircuitPython repo!</a>
-    </p>
-    <h2>Current Status for {{ site.data.libraries.updated_at | date: "%a, %b %d, %Y" }}</h2>
+<div id="libraries-page">
```
makermelissa, line 6: Same thing here.

makermelissa (changes requested):
Looks like you used the libraries-page ID multiple times (being in the layout and specific pages). Multiple duplicate IDs in the same page (because of it being in the common layout too) is something we should avoid. If both are supposed to be there, we should change it to use class instead. Also, maybe change it to something less page specific sounding?

jwcooper:
Need to make a couple more fixes quick.

tannewt (approved):
Looks good to me. Thank you!

makermelissa (approved):

makermelissa:
Thank you

PR #799, opened 2021-12-01
Rename manufacturer to remove single quote

jwcooper:
The quote causes issue with sorting. This was the quick fix.

Should be able to close out: https://github.com/adafruit/circuitpython-org/issues/524


makermelissa (approved):
Thanks for figuring that out.

PR #802, opened 2021-12-03
Add USB note

tannewt:


dhalbert (dismissed):
This is good -- I wonder if it should be made bold or otherwise called out.

dhalbert (approved):

apws:
ya, this is it

PR #803, opened 2021-12-03
Update images.yml per instructions

dhalbert:
Update image-actions per instructions in https://github.com/calibreapp/image-actions#usage.

The previous runs noted that the action was out of date.


dhalbert:
This note is visible on https://github.com/adafruit/circuitpython-org/pull/802, and is what provoked this PR:

image

makermelissa (approved):

PR #805, opened 2021-12-07
Add a buncha boards and add aliasing feature

makermelissa:
This brings the board count from 254 to 256 by adding variants such as the Pink Feather RP2040, which was accomplished by adding an alias feature. It also brings the blinka boards from 77 to 86 because of the U2IF support.


ladyada:
mm im a little unsure about the pink feather rp2040 separate - especially as soon we will only have pink (switching out black)

makermelissa:
I figured since we had variations already (such as https://circuitpython.org/downloads?q=Circuit+Playground) and it's a different PID, I would stick with the theme. I can remove it though.

jpconstantineau:
Will the aliasing feature be documented? I have been trying to figure out how it works and see if it matches my case.
Pykey60 is a keyboard with 61 keys that runs a RP2040.  Other keyboards I created uses the exact same CP build and pin definition but uses a subset of the rows/columns for the keyboard.  As such, I was planning to have a single build but for multiple keyboards.

Will the aliasing feature of cp.org allow multiple pages to document the board but all point to the same CP build?

As you can see, my use case is slightly different as the build can run on multiple boards, all a subset of the largest configuration.  This is quite different than the same hardware config but a different color PCB.

makermelissa:
> Will the aliasing feature be documented? I have been trying to figure out how it works and see if it matches my case. Pykey60 is a keyboard with 61 keys that runs a RP2040. Other keyboards I created uses the exact same CP build and pin definition but uses a subset of the rows/columns for the keyboard. As such, I was planning to have a single build but for multiple keyboards.
> 
> Will the aliasing feature of cp.org allow multiple pages to document the different boards but all point to the same CP build?
> 
> As you can see, my use case is slightly different as the build can run on multiple boards, all a subset of the largest configuration. This is quite different than the same hardware config but a different color PCB.

Yeah, the alias feature basically tells the board you are adding it to which board it should give you a cp build for. As long as a specific build will run on the other board it should be fine. For instance the PyPortal and PyPortal Pynt are the same build, but the aliasing was done in CircuitPython so they appear to be different builds and would have worked fine here.

I'll probably add a section to the guide on adding boards to circuitpython.org.

tannewt:
Generally, I'd just have different builds. It'll be clearer what build someone has if `board` only has the relevant pins.

tannewt (approved):
Looks good to me! I want to see the alias used very sparingly. Builds will all have the same board id even if the boards are different.

makermelissa:
> I want to see the alias used very sparingly.
I agree. :)

jpconstantineau:
@tannewt Agreed on the occasional use of this.  I guess it's a matter of keeping a fine balance between having lots of VID/PID requests to projects that allocate USB Product ID for Free Software / Open Hardware **__plus__** different CP builds __vs__ different pages in CP.Org.

These projects do push back when requests of multiple PIDs are made; especially when it's all for the same firmware.  After all, these "open" PIDs are in finite supply...

Openmoko commented the following in this [PR](https://github.com/openmoko/openmoko-usb-oui/pull/15#issuecomment-671754740):  _Since this is about a single firmware providing the same functionality, one USB ID should be enough. You can still figure out the actual board using USB strings or other USB configuration details._

More importantly, we wouldn't want people to add aliases to the Pi Pico to a board that uses a Pi Pico as the controller.  (same for all those keyboards that could potentially use the KB2040)

makermelissa:
> @tannewt Agreed on the occasional use of this. I guess it's a matter of keeping a fine balance between having lots of VID/PID requests to projects that allocate USB Product ID for Free Software / Open Hardware ****plus**** different CP builds **vs** different pages in CP.Org.
> 
> These projects do push back when requests of multiple PIDs are made; especially when it's all for the same firmware. After all, these "open" PIDs are in finite supply...
> 
> Openmoko commented the following in this [PR](https://github.com/openmoko/openmoko-usb-oui/pull/15#issuecomment-671754740): _Since this is about a single firmware providing the same functionality, one USB ID should be enough. You can still figure out the actual board using USB strings or other USB configuration details._
> 
> More importantly, we wouldn't want people to add aliases to the Pi Pico to a board that uses a Pi Pico as the controller. (same for all those keyboards that could potentially use the KB2040)

I think you can currently alias in CP without additional VID/PIDs. I think you just add it here: https://github.com/adafruit/circuitpython/blob/main/docs/shared_bindings_matrix.py

jpconstantineau:
@makermelissa Thanks for the link.  I'll need to look into this a whole lot more...  I suspect the docs on this would be relatively thin... I will probably need to ask for more...

@tannewt Any guidance on why one would prefer aliases here (CP.ORG) vs there (CP build)?

makermelissa:
It's mostly a convenience thing here either for boards that are electrically identical or if we wanted an alias and didn't want to wait for a new release of CP.

tannewt:
@jpconstantineau I think having different board ids is valuable. Aliases circumvent that. So, in general, create separate boards for things. We already need to optimize for many boards so having more is fine. To reuse a PID you can add to the whitelist here: https://github.com/adafruit/circuitpython/blob/main/tools/ci_check_duplicate_usb_vid_pid.py#L32

If you are selling your boards, you should really consider purchasing a VID. It'll give you 65k PIDs and you won't burden the open source pools that way.

jpconstantineau:
Here is the guideline the way I understand it:

1 - if a user of your board might have the need to be able to differentiate the board variant using `print(board.board_id)` then use CP Build alias 
2 - (else) if the variant of your board is sufficiently different that it needs separate documentation and a separate CP.ORG download page but not sufficiently different to have it's own build/board_id, then use the CP.ORG alias feature.

Most cases would fall under the first group...

tannewt:
Ya, I'm skeptical 2 would exist without 1.

PR #807, opened 2021-12-07
Add two Esp32s3 bringing the total to 256.

jepler:
image


ladyada (approved):
towards 0b1000000000000!

PR #808, opened 2021-12-08
Added page for imminent 2MB version of the Pimoroni Tiny 2040

ZodiusInfuser:
Also updated current Tiny 2040 page to clarify it is the 8MB version.

CircuitPython PR has only just been raised (see: https://github.com/adafruit/circuitpython/pull/5684) but getting this in now due to time zones.


makermelissa (approved):
Thanks

PR #814, opened 2021-12-14
Remove 6.x bundle links; add pointer to obsolete bundles list

dhalbert:
6.x bundle building was stopped by https://github.com/adafruit/circuitpython-build-tools/pull/83. This removes explicit listing of the 6.x bundles, and adds links to the last bundle built for a particular version.


makermelissa (approved):
Looks good.

PR #825, opened 2021-12-28
adding bluemicro833 board

jpconstantineau:
Adding board description as well as pictures of BlueMicro 833 board


jpconstantineau:
Just like the esp32 boards, I thought we should create a nRF52833 family to differentiate the boards in the filter.
Once this is in, I can go back and clean up the other 833 boards and switch their family.

jpconstantineau:
On second thought, I think a more detailed discussion on the family/chip is needed...

As of now, the following families are in place:

- Atmel-Samd
- Cxd56
- Esp32c3
- Esp32s2
- Esp32s3
- Litex
- Mimxrt10xx
- Nrf52840
- Raspberrypi
- Stm

However, if we really want to make this right, we would need to change "nrf52840" to "nRF52" since the nrf52840 is a specific chip, not a whole family.  If we want to get down to the chip details, we would need to have SAMD21 and SAMD51 as this is likely a split that would be quite handy to have.  Ultimately, we would want to be able to filter by both levels (see below) but that would need further work on the page itself...

- Atmel-Samd
   - SAMD21
   - SAMD51
- Cxd56
- Esp32c3
- Esp32s2
- Esp32s3
- Litex
- Mimxrt10xx
- nRF52
   - nRF52833
   - nRF52840
- Raspberrypi
   - RP2040 
- Stm
   - STM32L072
   - STM32F405
   - STM32F411
   - STM32F746

makermelissa (approved):
Looks good. Thanks!

makermelissa:
I agree on the discussion, but think that should be a separate PR.

PR #828, opened 2021-12-29
Remove RP2040 copy-paste from description.

dgriswo:
Looks like a copy/paste miss.  It's the Feather RP2040 product page on the Feather ESP32-S2 w/BME280 sensor board.


jepler (approved):
Thank you, nice catch. I fixed the text up slightly.

PR #833, opened 2022-01-05
Get some straggling 'contribute' headers removed

jepler:
These had three hash marks before "Contribute". Other headings in these files may also have had "###" where "##" is usual, further review wouldn't hurt.


makermelissa (approved):
Thanks

PR #845, opened 2022-01-15
Update esp32s3 devkits

prplz:
For https://github.com/adafruit/circuitpython/pull/5848.


tannewt (approved):
@makermelissa Are you ok with these three boards sharing images?

makermelissa:
Yeah, I'm fine with that.

PR #847, opened 2022-01-19
[typo] SPI pins are not puns

2bndy5:
I'm guessing you meant "high speed SPI peripheral pins".

2 jokes come to mind:

1. Does "high speed puns" mean people don't have to think as much when I try to make a pun?
2. I think your finger had a vowel movement.


FoamyGuy (approved):
Looks good to me. Thanks for the fix @2bndy5

makermelissa:
Good catch. This was copy/pasted from the product page and it's on there too.

2bndy5:
I was hoping that fixing it here would also be reflected on the product page.

ladyada:
nope, different CMS! fixed now

PR #850, opened 2022-01-24
Update adafruit_feather_rp2040.md

sctv:
Updated that rp2040 based feather can use an Arduino core.  Welcome feedback on how to phrase it better.
A similar line should be included in all rp2040 based boards. 
Thanks.


jepler:
Thanks, this would be a good update to make.

Because we like the text for Adafruit products to match what's in the shop, I've raised this issue internally. I'll get back to you about this, and if I don't you should feel free to ping me about it in a few days.

makermelissa (changes requested):
It looks like you have some duplicate text in there. Want to fix it?

sctv:
@jepler @makermelissa will sync this description up with the product page [rp2040 feather](https://www.adafruit.com/product/4884) tomorrow. Still getting my head around git/hub/branches/pr's etc.
thanks for your patience. Have allowed edits by maintainers on patch-2.

makermelissa (approved):
Looks good. Thanks for making the changes. Also we prefer to assist others in learning how to use GitHub instead of just doing it for them.

PR #872, opened 2022-02-24
Add LILYGO TTGO T8 ESP32-S2-WROOM

fabaff:
Depends on https://github.com/adafruit/circuitpython/pull/6084


makermelissa (approved):
Looks good and the dependent PR is merged now.

PR #876, opened 2022-02-24
Compress features list and add links to details

fabaff:
The feature list is now a bit shorter. The details were a bit extensive from my point of view and mostly covering the ESP32-S2FN4R2 and not so much the S2 Pico itself.


tannewt (approved):
Thank you!

PR #884, opened 2022-03-01
Reduce content and update specs

fabaff:
Remove "marketing content" and try to stay on point. Remove repetitions.


makermelissa (approved):
Thanks

Neradoc:
Why are you editorializing the description of the other people's boards ?

dhalbert:
Making the headings consistent is good, but I agree that trimming content is something we should ask of the original submitters. It was OK with us before. I would suggest reverting this and making the consistency changes only for now.

PR #890, opened 2022-03-11
Additional Notes for the Itaca uChip

mew-cx:
A couple of notes from my experience with the Itaca uChip.

### DIFF
--- a/_board/uchip.md
+++ b/_board/uchip.md
@@ -47,6 +47,12 @@ You can program uChip using virtually any IDE, in many languages (CircuitPython 
 * Standards narrow-DIP footprint: 0.3” (7.62 mm) row spacing, 0.1” (2.54) pin spacing.
 * Pinout standard logic CMOS compatible: power and GND are on pin 16 and 8, so you can also emulate some 16 pin CMOS ICs (4000 and 74HC series)!  
 
+**Additional Notes:**
+
+* The uChip does not have an RGB LED.
+* The on-board LED is amber, rather than green.
+* The board's bootloader is BOSSA-compatible, but does not support UF2, so use a CircuitPython .bin file with BOSSA.  See [this page](https://learn.adafruit.com/welcome-to-circuitpython/non-uf2-installation) for the general process of using BOSSA.  A Windows .bat script that demonstrates exactly how to use the BOSSAC commandline tool with uChip to flash CircuitPython is available [here](https://github.com/mew-cx/circuitpython_stuff/tree/main/uChip); you may use it as a template and customize as needed.
+
 ## Purchase
 
 * [Itaca Innovation Store](https://shop.itaca-innovation.com)


_board/uchip.md
```diff
@@ -47,6 +47,13 @@ You can program uChip using virtually any IDE, in many languages (CircuitPython
 * Standards narrow-DIP footprint: 0.3” (7.62 mm) row spacing, 0.1” (2.54) pin spacing.
 * Pinout standard logic CMOS compatible: power and GND are on pin 16 and 8, so you can also emulate some 16 pin CMOS ICs (4000 and 74HC series)!  
 
+**Additional Notes:**
+
+* The uChip does not have an RGB LED.
```
fabaff, line 52: IMHO, we shouldn't mention hardware elements that are missing. The "features list" says that there is a LED.

_board/uchip.md
```diff
@@ -47,6 +47,13 @@ You can program uChip using virtually any IDE, in many languages (CircuitPython
...
+* The on-board LED is amber, rather than green.
```
fabaff, line 53: I don't get the relevance here. For most use cases the color of the status LED doesn't matter.

_board/uchip.md
```diff
@@ -47,6 +47,13 @@ You can program uChip using virtually any IDE, in many languages (CircuitPython
...
+* The board's bootloader is BOSSA-compatible, but does not support UF2.  Use the .bin image of CircuitPython with BOSSA.
```
fabaff, line 54:
```suggestion
* The board's bootloader is BOSSA-compatible, but does not support UF2. Use the .bin image of CircuitPython with BOSSA.
```

mew-cx:
Let me explain my rationale for these proposed changes.

Since the _vast_ majority of CircuitPython, its docs, and supporting boards are products of Adafruit, I feel it would be useful to describe how this board differs from common expectations of those Adafruit boards, to prevent accidental assumptions.  Ex: if a project for some reason requires green light, it could be handy to know that isn't included, not even via an RGB LED; AFAIK that LED color is not actually documented anywhere, not even on Itaca's site.

_board/uchip.md
mew-cx: see previous comment

_board/uchip.md
mew-cx: see previous comment

mew-cx: The color was mentioned in case it _does_ matter.

dhalbert: Mentioning this is fine with us.

dhalbert:
@mew-cx: To clarify, @fabaff is an independent reviewer, and is not working for Adafruit.

@fabaff we appreciate your reviews and PR's. We may differ on policy questions.

_board/uchip.md
```diff
@@ -47,6 +47,13 @@ You can program uChip using virtually any IDE, in many languages (CircuitPython
...
+* I've created a Windows .bat script using the BOSSAC commandline tool to flash CircuitPython; see https://github.com/mew-cx/circuitpython_stuff/tree/main/uChip Please customize as needed.
```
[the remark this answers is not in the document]
mew-cx: > The collected details add value.

Indeed, that's the point: _all_ details are captured in a self-documented script; even the filename describes the specifics.  It serves as a single-point-of-truth fully-functional template available for tweaking.  I am my own customer (I use the script myself, tweaking as needed), rather than creating a potentially out-of-sync description for others.

> the binary is in $PATH.

The BOSSA installer does not modify PATH.

> (Sorry, I'm not a Windows users):

Neither am I, really.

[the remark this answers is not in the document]
dhalbert: @mew-cx We have a writeup of using bossac here: https://learn.adafruit.com/welcome-to-circuitpython/non-uf2-installation. I think the main point to make is that you need to use bossac: there is no UF2 bootloader (If I understand correctly). I think it would make sense to refer to the guide page, since it probably subsumes your more specific script.

dhalbert (changes requested):
Thanks! One request re bossac.

mew-cx:
> We have a writeup of using bossac here

Yep, I used that :-) that's how I created the script.  Agree RE adding a link.  However, I believe there's utility in a known-good script with _all_ the decisions resolved and documented (albeit for one usecase), rather than requiring every user to resolve from a description.  It's an idiom I've embraced over the years.

mew-cx:
edited and squashed.  TPAL

dhalbert (approved):
OK, looks good to me. Thanks!

PR #894, opened 2022-03-14
This is ESP32-S3 not ESP32-S2

dglaude:
Just changing the obvious confusion between ESP32-S2 and ESP32-S3...

But maybe the first YouTube video need to be removed, it does not seems to talk about ESP32-S3 but only about ESP32 and playing doom or game emulation.


makermelissa (approved):
Thanks. Looks like this needs to be updated with product info, but I can do that in another PR.

PR #896, opened 2022-03-14
Update Adafruit QT Py ESP32-C3

makermelissa:


ladyada (changes requested):
you edited teh esp32s3 not c3

makermelissa:
😳 Oops

makermelissa:
It doesn't looks like this board has been added to CircuitPython yet and I can't find anything new on the S3, so I'll just close this and stash away the images for a later date.

PR #904, opened 2022-03-22
Remove reference to TTGO T8 ESP32-S2

fabaff:
TTGO T8 ESP32-S2 is now also [supported](https://circuitpython.org/board/lilygo_ttgo_t8_s2/).


makermelissa (approved):
This looks ok.

PR #923, opened 2022-04-01
Added board page for new Pimoroni Servo 2040

ZodiusInfuser:
This launched today, and already has the CircuitPython definition merged into the main release.
https://shop.pimoroni.com/products/servo-2040


makermelissa (approved):
Thanks!

PR #924, opened 2022-04-02
Added missing boards plus new blinka board

makermelissa:


dhalbert (approved):
Thanks for all this work!

PR #935, opened 2022-04-22
Fix lilygo_ttgo_t8_esp32_s2_wroom issues

### LINKED ISSUE #933: lilygo_ttgo_t8_esp32_s2_wroom: Mismatch between URL and board ID.
In [lilygo_ttgo_esp32_s2_wroom.md](https://github.com/adafruit/circuitpython-org/blob/main/_board/lilygo_ttgo_esp32_s2_wroom.md), the board ID is `lilygo_ttgo_t8_esp32_s2_wroom`. This matches the download filenames and [the board name in the circuitpython repo](https://github.com/adafruit/circuitpython/tree/main/ports/espressif/boards/lilygo_ttgo_t8_esp32_s2_wroom). However, the markdown filename and therefore the URL is `lilygo_ttgo_esp32_s2_wroom` (i.e. without `t8_`). This results in the "Edit the source for this page" link pointing to the wrong file.

The link in the right hand pane to download the bootloader is also wrong. It points to https://github.com/adafruit/tinyuf2/releases/download/0.8.0/tinyuf2-lilygo_ttgo_t8_s2_wroom-0.8.0.zip which doesn't exist. The only one I can find is https://github.com/adafruit/tinyuf2/releases/download/0.9.0/tinyuf2-lilygo_ttgo_t8_s2_wroom-0.9.0.zip for version `0.9.0`.

makermelissa:
Fixes #933.


dhalbert (approved):
Thanks - I was meaning to do this.

PR #942, opened 2022-05-03
Fix 2 unknown boards

makermelissa:


FoamyGuy (approved):
Changes look good to me. 

Tested with a locally running server. The micromod device now shows images and more information and the Motor 2040 has been removed from the downloads list for now which was the intention as far as I understand it.

PR #944, opened 2022-05-05
Add missing atmel-samd and nrf52840 bootloader IDs

### LINKED ISSUE #943: Bootloader section missing for some boards
For example, not showing up for Matrix Portal:

image

makermelissa:
Fixes #943.

### DIFF
--- a/_board/ndgarage_ndbit6_v2.md
+++ b/_board/ndgarage_ndbit6_v2.md
@@ -9,6 +9,8 @@ board_image: "unknown.jpg"
 downloads_display: false
 date_added: 2020-09-1
 family: atmel-samd
+bootloader_id: ndbit7
+
 ---
 
 Oops! Looks like we don't know anything about this board. This means it's likely very new.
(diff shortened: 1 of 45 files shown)


_board/ndgarage_ndbit6_v2.md
```diff
@@ -9,6 +9,8 @@ board_image: "unknown.jpg"
 downloads_display: false
 date_added: 2020-09-1
 family: atmel-samd
+bootloader_id: ndbit7
```
kattni, line 12: Is this supposed to be `ncbit6`?

kattni (changes requested):
Looks good! Caught one thing below.

_board/ndgarage_ndbit6_v2.md
makermelissa: I don't believe so. There was already an ndbit6 board (different than this) and there is nothing called ndbit7, so I figured they were being clever.

kattni: Fair enough!

kattni (approved):

PR #945, opened 2022-05-06
Add missing espressif bootloader IDs

### LINKED ISSUE #943: Bootloader section missing for some boards

makermelissa:
Fixes #943. Pulled IDs from here: https://github.com/adafruit/tinyuf2/releases. Some were missing. A couple C3 bootloader IDs removed due to not using native USB.

### DIFF
--- a/_board/adafruit_qtpy_esp32s2.md
+++ b/_board/adafruit_qtpy_esp32s2.md
@@ -8,6 +8,7 @@ board_url: "https://www.adafruit.com/product/5325"
 board_image: "adafruit_qtpy_esp32s2.jpg"
 date_added: 2021-11-30
 family: esp32s2
+bootloader_id: adafruit_qtpy_esp32s2
 features:
   - STEMMA QT/QWIIC
   - USB-C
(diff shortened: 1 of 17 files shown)


_data/bootloaders.json
```diff
@@ -9,6 +9,12 @@
         "esp32s2": {
             "version": "0.9.0"
         },
+        "esp32s3": {
+            "version": "0.9.0"
+        },
+        "esp32c3": {
+            "version": "0.9.0"
+        },
```
Neradoc, line 17:
```suggestion
        "esp32c3": {},
```
Since there is no build of tinyuf2 for C3, we should leave it empty.

Neradoc (changes requested):
Thanks !
One point in bootloaders.json, and I think the QT PY ESP32C3 needs its bootloader_id removed too.
https://github.com/adafruit/circuitpython-org/edit/main/_board/adafruit_qtpy_esp32c3.md

_data/bootloaders.json
makermelissa: Good call.

makermelissa:
Updated. Anything else?

Neradoc (approved):
Seems all good to me. The hiibot iots2 has been added to the tinyuf2 CI, but it won't be available until the next release.

PR #947, opened 2022-05-10
Check features at build time

jepler:
Before this, people were requested to follow a set list of features, but we had no automated check for it. Do this as a step of the CI, or a step that can be run locally.

I found several boards that had non-standard features listed; I corrected or deleted them as I saw appropriate.

There are also a number of boards without any features: blocks. I did not attempt to fix this. However, if this were fixed we could add an additional check in this script for future boards. However, this assumes every board has at least one feature in the list of features, which may not be true.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -30,6 +30,10 @@ jobs:
     - uses: actions/checkout@v1
       with:
         submodules: true
+    - name: Set up Python 3.x
+      uses: actions/setup-python@v2
+      with:
+        python-version: 3.x
     - uses: ruby/setup-ruby@v1
       with:
         ruby-version: 3.0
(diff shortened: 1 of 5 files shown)


jepler:
Here's an example of the check failing when a nonstandard feature is specified:
```
jepler@rat:~/src/circuitpython-org$ git diff
diff --git a/_board/feather_m4_express.md b/_board/feather_m4_express.md
index cdde3923..b8948476 100644
--- a/_board/feather_m4_express.md
+++ b/_board/feather_m4_express.md
@@ -10,6 +10,7 @@ date_added: 2019-3-8
 family: atmel-samd
 bootloader_id: feather_m4
 features:
+  - OSHWA Certified
   - Feather-Compatible
   - Battery Charging
   - Breadboard-Friendly
jepler@rat:~/src/circuitpython-org$ ./check-features.py 
_board/feather_m4_express.md:0: Non-standard feature: OSHWA Certified
```

The format of the message is to make it match the convention used in many standard programs and recognized by standard editors like vi and emacs so that the file containing the error can be easily visited. ":0:" means that the line number where the problem occurred is not known.

PR #948, opened 2022-05-10
Update Check Features to check Blinka boards too

### LINKED ISSUE #583: Loop through a set of known features only
I propose a change to the downloads page (boards and Blinka) that instead of showing every different variation of features in the download set (in which it is easy to miss something), we show a defined list of features and if something doesn't match, we don't show it.

makermelissa:
Fixes #583.


jepler (approved):
Thank you!

PR #949, opened 2022-05-11
List frozen modules in the download page

Neradoc:
This lists the frozen modules below the built-in ones on boards pages.
Also includes a fix of files.json to replace frozen dir names with actual module names, [as mentioned here](https://github.com/adafruit/circuitpython-org/issues/684#issuecomment-1100794905).

The (?) link points to [the glossary's short paragraph on frozen modules](https://docs.circuitpython.org/en/latest/docs/reference/glossary.html?highlight=frozen#term-frozen-module).
Another possibility was [the frozen modules page in the "building circuitpython" guide](https://learn.adafruit.com/building-circuitpython/adding-frozen-modules).

It would be nice to have a more end-user-oriented reference on frozen modules, explaining that they take precedence over the lib directory, and how to replace them with a newer version (by placing a new version in the root directory). Maybe as a page in docs.circuitpython.org.

Capture d’écran 2022-05-11 à 13 53 34

### DIFF
--- a/_data/files.json
+++ b/_data/files.json
@@ -8,8 +8,8 @@
      "uf2"
     ],
     "frozen_libraries": [
-     "Adafruit_CircuitPython_HID",
-     "Adafruit_CircuitPython_SD"
+     "adafruit_hid",
+     "adafruit_sdcard"
     ],
     "languages": [
      "de_DE",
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
This looks great. Thanks for making these changes.

PR #950, opened 2022-05-12
Add "Xiao / QTPy Form Factor" feature

jepler:
Is this the best 'spelling' for the feature? Is it worth adding at all?

Closes adafruit/circuitpython#6371

Also updates awesome-circuitpython to latest


makermelissa (approved):
Sounds good. I'll add to the guide.

PR #952, opened 2022-05-12
More Xiaos that have the Xiao form factor

Neradoc:
Also I didn't go through the features they could have, but at least made "breadboard-friendly" consistent.


makermelissa (approved):
Looks good. Thanks.

PR #959, opened 2022-05-19
Add IoTS2 firmware ID and bump tinyuf2 version

Neradoc:
`tinyuf2` 0.9.2 was released, adding the Hiibot IoTS2 files to the assets.


makermelissa (approved):
Awesome. Thanks for adding.

PR #962, opened 2022-05-24
Contribute section seems automatic...

dglaude:
The automatic version goes to `main` where the manual version was going to `master`
So it appear duplicate in https://circuitpython.org/board/mixgo_ce_serial/

Maybe it is not clearly documented that this section is not to be submitted manually?

### DIFF
--- a/_board/mixgo_ce_serial.md
+++ b/_board/mixgo_ce_serial.md
@@ -29,5 +29,5 @@ Mixgo CE Serial(Serial port loader) is designed by mixly team and can be program
 * 4-way Type-C USB external expansion
 
 ## About
-* [Milxy](https://mixly.org/)
+* [Mixly](https://mixly.org/)
 * [Mixly online programming Edition](https://xmote.org/)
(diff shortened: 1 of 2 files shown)


dhalbert (approved):
Thanks for noticing this! If you have a suggestion for the README about the auto-"Contribute" section", a PR there would be welcome as well.

dglaude:
I did not check the documentation, but I checked 20+ of the latest new board, and only two had that duplicate, and from the same author. So I guess the instruction are OK and the same mistake was done twice by the same author.
Maybe the "Contribute" section could be put elsewhere, more global to the page and not in the same text box, so that what appear in the box is exactly and only what was provided in the md file for this board.

dahanzimin:
Thanks for helping to correct some errors. When can I download mixgo_ce firmware from the circuitpython-org

makermelissa:
The button should appear in the next release of CircuitPython, but you can click the Browse S3 button now to see all builds which include any that happen when a Pull Request gets merged.

makermelissa:
#965 will automatically check for this from now on.

PR #963, opened 2022-05-24
Contribute section seems automatic...

dglaude:
The automatic version goes to `main` where the manual version was going to `master`
So it appear duplicate.

Maybe it is not clearly documented that this section is not to be submitted manually?


dhalbert:
DIdn't realize you made two PR's. Repeated this in my updates to #962, so closing this in favor of #962.

PR #965, opened 2022-05-26
Added additional checks for boards and images

### LINKED ISSUE #957: Add Additional PR Checks
Now that we are checking the features, some additional checks that would be nice for PRs to have are:

- [x] A "date_added" field is present and has a valid value
- [x] A "family" field is present and of a supported type
- [x] Image sizes are appropriate
- [x] The image in the board file is present
- [x] If board is set to not display due to lack of info, then maybe skip some of these checks
- [x] No contribute Section includes

That should cover the most common things I tend to see that I would request changes for or update at a later date once I notice the info is missing.

makermelissa:
Fixes #957. This also corrects a number of things found during testing such as image sizes and missing information.


_board/diodes_delight_piunora.md
```diff
@@ -8,6 +8,7 @@ board_url: "https://www.crowdsupply.com/diodes-delight/piunora"
 board_image: "diodes_delight_piunora.jpg"
 download_instructions: 
 date_added: 2021-2-14
+family: raspberrypi
```
tannewt, line 11: Want to call this broadcom to distinguish from rp2040? Thats what we do it CP.

tannewt (changes requested):
This looks really handy! I think you'll want to split raspberrypi (rp2040) from broadcom (raspberry pi SBC boards) though.

_board/diodes_delight_piunora.md
makermelissa: Sure

makermelissa:
Yep, thanks. They're now split. I'm hoping it will save us a lot of manual work and frustration.

tannewt (approved):
Thank you!

PR #967, opened 2022-05-28
Always update awesome-circuitpython during build

jepler:
Closes: #880


dhalbert (approved):
Looks good, and I double-checked the adabot version.

jepler:
Oh, I didn't intentionally update adabot so thank you for double-checking that!

PR #968, opened 2022-06-01
Display issue labels for open issues

### LINKED ISSUE #646: Add issue label to the "/contributing" page
When searching for an issue, or trying to get an overall picture of the issues list, it would be nice to see, at a glance, the labels for an issue. That could be either by grouping issues together by their label, or by displaying the label(s) of an issue next to it, similar to the issue's age.

makermelissa:
Fixes #646. I attempted to use `:after` style css labels first, but css spec only allows 1 `:after` per item, so this was the best approach to display multiple tags.

### DIFF
--- a/assets/sass/pages/_contributing.scss
+++ b/assets/sass/pages/_contributing.scss
@@ -23,4 +23,60 @@
       }
     }
   }
+
+  ul.issues-list{
+    li {
+      .issue-label {
+        border-radius: 2em;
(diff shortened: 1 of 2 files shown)


makermelissa:
Screen Shot 2022-06-01 at 10 00 00 AM

tannewt (approved):
Very cool! Thank you!

PR #975, opened 2022-06-09
Add newest Feather ESP32-S3

makermelissa:


dhalbert (approved):
Looks good! I added this board to CPy, so I'm familiar with it.

PR #980, opened 2022-06-16
Fix Adafruit Feather ESP32S2 bootloader id and bump tinyuf2

Neradoc:
Fixes the bootloader_id for Feather ESP32-S3 4MB Flash 2MB PSRAM.
Bump tinyuf2 from 0.9.2 to 0.9.4 as required.

https://github.com/adafruit/tinyuf2/releases/tag/0.9.3


ladyada (approved):

ladyada:
melissa might be out for a while in which case im fine with approving based on "it doesnt seem like it could make anyhting worse" :)

PR #985, opened 2022-07-05
Add Wemos C3 Board

askpatrickw:
Fixes #984

PR #986, opened 2022-07-06
Update checker script to fail if images are missing

makermelissa:
This change will catch incorrect file names like in #985.


jepler (approved):
Thank you!

PR #987, opened 2022-07-07
Seeed Studio is making a naming change on the XIAO serie

dglaude:
Seeed Studio is making a naming change on the XIAO serie of board, so I adapted 3 boards (skipping "Seeeduino XIAO - Keyboard Optimized")

https://www.seeedstudio.com/blog/2022/07/06/notice-on-product-name-change/

### DIFF
--- a/_board/Seeed_XIAO_nRF52840_Sense.md
+++ b/_board/Seeed_XIAO_nRF52840_Sense.md
@@ -1,8 +1,8 @@
 ---
 layout: download
 board_id: "Seeed_XIAO_nRF52840_Sense"
-title: "Seeed XIAO nRF52840 Sense Download"
-name: "Seeed XIAO nRF52840 Sense"
+title: "Seeed Studio XIAO nRF52840 Sense Download"
+name: "Seeed Studio XIAO nRF52840 Sense"
 manufacturer: "SEEED"
 board_url: "https://www.seeedstudio.com/Seeed-XIAO-BLE-Sense-nRF52840-p-5253.html"
 board_image: "Seeed_XIAO_nRF52840_Sense.jpg"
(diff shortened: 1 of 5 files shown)


Neradoc:
Makes sense to me, maybe add something like "(Formerly Seeeduino Xiao)" for search purposes ?

makermelissa:
Would you like to updated the Manufacturer to "Seeed Studio" as well?

makermelissa (changes requested):
Waiting on response to Neradoc's and my questions...

ladyada:
@dglaude please check the questions above

dglaude:
> Seeed Studio

If `SEEED` was their old name and `Seeed Studio` the new name, so yes that make sense.
But I don't know how to do that (it is used in the filter) and it also affect the Wio Terminal, so 5 boards in total.

PS:
I am not working for SEEED/Seeed Studio. Maybe those that contribued those board in the first place can propose a fix. :-)
This can be a separate PR, I only noticed the XIAO naming issue and that is what I tried to fix here.

makermelissa:
Hi, no worries. I wasn't assuming you worked for them. 

To change the name, just find the boards: https://circuitpython.org/downloads?manufacturers=SEEED (and it looks like you already located them), then in the board file, change manufacturer to "Seeed Studio". As long as you update all of the boards and they have the exact same spelling, the filter will update automatically (otherwise there will be a filter for every variation).

dglaude:
> change manufacturer to "Seeed Studio"

Hi @makermelissa I changed the 5 boards.

If that company want to improve those descriptions further or submit more board (I am pretty sure they have some more), I suggest to let them do it. I don't want me nor you to do that work for them. ;-)

makermelissa (approved):
Looks good. Thank you.

PR #1000, opened 2022-07-21
Update subheader/brief description of board

kylefmohr:
The live version says this has an OLED, that's not true. I made minor modifications to Wemos's own description of the board [here](https://www.wemos.cc/en/latest/c3/c3_mini.html)


makermelissa (approved):
Thanks

PR #1005, opened 2022-07-26
Improve ESP32-nn bootloader installation instructions

### LINKED ISSUE #1001: ESP32-S3: Add information on using the UF2 bootloader
The ESP32-S3 board descriptions explain how to install the UF2 bootloader.

I recommend adding details on how to trigger into UF2 bootloader mode:

1. Click RESET
2. While purple LED is lit, click BOOT. (LED will flash red and then remain green.)
3. `S3DKC1BOOT` drive should appear

Also, it may be helpful to indicate that you should connect to the UART port to flash the UF2 boot loader. And then after that is complete that you then connect to the USB port when using the UF2 boot loader to get to the S3….BOOT drive or the CIRCUITPy drive

dhalbert:
Fixes #1001.

Caution users to refer to board documentation to update or install UF2 bootloader.
Describe how to enter UF2 bootloader.

### DIFF
--- a/_includes/download/board.html
+++ b/_includes/download/board.html
@@ -197,14 +197,17 @@ By the way, boolean operation precedence is right to left! (yeesh)
   </p>
 
   <p>
-    <em>Note: <b>update.uf2</b> files are not currently working on ESP32-S2 boards.</em>
+    <em>Note: <b>update.uf2</b> files are not currently working on ESP32-S2 or ESP32-S3 boards.</em>
   </p>
 
   <p><strong><em>Important</em></strong>:
     <em>this will erase previously flashed firmware and sketches from the board,
       but needs to be perfomed only once.</em>
   </p>
-
+  <p><em>The instructions here are general.
+      We recommend you consult the manufacturer's board documentation for detailed
+      instructions, which may be different.</em>
+  </p>
  <ul>
   <li>Unzip to find the file <b>combined.bin</b>.</li>
   <li>Place board in bootloader mode:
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thanks. Looks good.

kmatch98:
The update may cover some ESP32-XX boards, but the ESP32-S3 devkits are slightly different than the instructions you added. 

The ESP32-S3 devkit boards have both a RESET and a BOOT button. Also, this board has “UART” and “USB” ports. 

After installing the UF2 bootloader using the “UART” port, now connect your PC to the “USB” port. 
To enter the UF2 bootloader mode, click the RESET button.  As soon as the LED turns purple, click the BOOT button.  The LED should quickly flash red and the stay steady green, and the `S3DKC1BOOT` should appear.

Neradoc:
Yeah, almost only Adafruit boards have the double reset feature. It's a support issue that comes up. And most boards are not Adafruit boards anymore, and will not have a learn guide or instructions about using tinyUF2. I would prefer if instructions with boot were given.

> After installing the UF2 bootloader using the “UART” port
It works on the USB port too. You can completely ignore the UART port on the devkit, it's an additional CP2102 connected to the debug UART.

dhalbert:
I don't really want to embed detailed instructions in this circuitpython.org page, because it's harder to maintain than a Learn Guide page, and the instructions may not be applicable to all ESP32-nn UF2-capable boards. That is what I was trying to convey as a caveat. I think it would better to have general-purpose Learn Guide page for these non-Adafruit boards that we can link to. I was thinking of removing the reference to the "ESP Web Flasher", for instance.

dglaude:
I understand @dhalbert but my practice is to go to circuitpython.org download and get a hit from there on how to upgrade...

The first reason for having the information on that page is to know if I have to download the ".bin" or the ".uf2" (else I will download both, just in case).

The issue and pain is when working with non Adafruit board or unusual Adafruit board that run CP such as STM32, Teensy, or that have special button clicking like the Wio Terminal, I fail remember and I hate to redo the same search, it just mean I will upgrade those less frequently, or avoid using them if possible.
So if basic information is not on that page, it should be one click away somewhere.
Any tips and trick part of the (initial install and) upgrade should be there, such as using CLI to go in UF2 mode at next boot.

Anything to make solution that help frequent updater that fait to remember would be welcome.

PR #1006, opened 2022-08-01
Create webworkflow.css

andywarburton:
We want to use remote CSS pulled from the main circuitpython.org site to enhance the usability of Web Work Flow without significantly impacting firmware size. This is an initial file so I can test and make sure it works as intended.


tannewt (approved):
Thank you!

PR #1009, opened 2022-08-02
Wrong information how to enter UF2 bootloader

mikaelsundin:
Wrong infromation how to enter UF2 bootloader.

No propose changes button exist for "Install, Repair, or Update UF2 Bootloader" section on page 
https://circuitpython.org/board/lolin_s2_mini/

### DIFF
--- a/_board/lolin_s2_mini.md
+++ b/_board/lolin_s2_mini.md
@@ -72,6 +72,15 @@ The LOLIN S2 Mini is a small (33.4 mm x 25.4 mm) development board. The form fac
 * [Dimensions](https://www.wemos.cc/en/latest/_static/files/dim_s2_mini_v1.0.0.pdf)
 * [Factory MicroPython firmware](https://www.wemos.cc/en/latest/tutorials/s2/get_started_with_micropython_s2.html)
 
+## Enter bootloader
+
+To enter UF2 Bootloader Version 0.10.2:
+* Click reset button 
+* Wait until LED is lit
+* Hold BOOT button
+
+Be sure not to hold BOOT button when RESET button is pressed.
+
 ## Purchase
 
 * [AliExpress](https://www.aliexpress.com/item/1005003145192016.html)


makermelissa (changes requested):
Ok, so the problem with this approach is that the boot loader section will still be there, which could be confusing (See https://github.com/adafruit/circuitpython-org/blob/main/_includes/download/board.html#L171-L248). Also, since I don't have the Lolin S2 Mini, I can't verify that these steps are different, so I'll have to take your word on it. You could always remove the value from bootloader_id to hide the section, but the downloadable file wouldn't be there.

I'm wondering if there may be a better approach to this like specifying a `hide_bootloader_instructions` in https://raw.githubusercontent.com/adafruit/circuitpython-org/main/_board/lolin_s2_mini.md and then checking for that. 

Please let me know your thoughts on that based on your comfort level with making changes like this.

Neradoc:
> Ok, so the problem with this approach is that the boot loader section will still be there, which could be confusing (See https://github.com/adafruit/circuitpython-org/blob/main/_includes/download/board.html#L171-L248). 

Note that this is only one paragraph, the one added in #1005 

> Also, since I don't have the Lolin S2 Mini, I can't verify that these steps are different, so I'll have to take your word on it.

These are the generic instructions for any ESP32-Sx board. The instructions written as they are now are specifically for Adafruit boards, as I mentioned: https://github.com/adafruit/circuitpython-org/pull/1005#issuecomment-1196602500
We need a better solution than editing the board's file. Dan mentioned a general-purpose Learn Guide page for detailed instructions.

mikaelsundin:
The information is available in Adafruit Tinyuf2 gitrepo.

https://github.com/adafruit/tinyuf2/blob/master/ports/espressif/boards/lolin_s2_mini/board.h
Line 31

> // Enter UF2 mode if GPIO is pressed while 2nd stage bootloader indicator
> // is on e.g RGB = Purple. If it is GPIO0, user should not hold this while
> // reset since that will instead run the 1st stage ROM bootloader
> #define PIN_BUTTON_UF2        0

makermelissa:
> > Ok, so the problem with this approach is that the boot loader section will still be there, which could be confusing (See https://github.com/adafruit/circuitpython-org/blob/main/_includes/download/board.html#L171-L248).
> 
> Note that this is only one paragraph, the one added in #1005
> 
> > Also, since I don't have the Lolin S2 Mini, I can't verify that these steps are different, so I'll have to take your word on it.
> 
> These are the generic instructions for any ESP32-Sx board. The instructions written as they are now are specifically for Adafruit boards, as I mentioned: [#1005 (comment)](https://github.com/adafruit/circuitpython-org/pull/1005#issuecomment-1196602500) We need a better solution than editing the board's file. Dan mentioned a general-purpose Learn Guide page for detailed instructions.

Ok, I can go ahead and approve/merge this for now with the understanding that this will be better addressed later.

makermelissa (approved):
After a little discussion, this is fine.

PR #1018, opened 2022-08-15
Fixing case for microDev vs MicroDev

dglaude:
There are two board by "microDev" but one is spelled "MicroDev", so I decided on a single case:
* manufacturer: "microDev"

Otherwise you have two entry in the Filter part.


makermelissa (approved):
Thanks. Good Catch.

MicroDev1:
Nice Catch @dglaude! Thanks for spotting this, I have now started using "MicroDev"; I'll make a PR to update it.

PR #1019, opened 2022-08-15
Make sure all 3 board use "Switch Science, Inc"

dglaude:
3 board but two spelling:
* Two board use `manufacturer: "Switch Science, Inc"`
* One board use `manufacturer: "Switch Science, Inc."`

So I moved to the majority without that "."


makermelissa (approved):
Thanks.

PR #1022, opened 2022-08-19
Added Adafruit ESP32 boards

makermelissa:
These are the easiest boards to add, but there are still 13 more to do in another PR.


ladyada (approved):

dhalbert:
Thanks!!

PR #1024, opened 2022-08-21
Add Waveshare ESP32-S2-Pico

bill88t:
I'm tired of babysitting this board, let's get this over with.
Disclaimer, I am not affiliated in any way with Waveshare.

The stuff on the markdown are a few of the bulletpoints in their wiki and product page.
The photos were snapped by me, as the ones provided by waveshare are smol (<900px).

I do not have the display variant of this board, nor can I aquire it, so it will have to stay unkown for now.


makermelissa (approved):
Thank you so much for adding it.

PR #1026, opened 2022-08-22
Generate the S3 link for inkown boards using URL query string

### LINKED ISSUE #1025: Possible improvement for unknown boards page
Ok, I've had a seed of an idea for this for a while, but I hadn't thought of a good way to implement it until now. One big challenge with this website is that the pages are static, which makes displaying information that doesn't exist challenging. The only thing we do know is the board id on the page that shows all of the boards.

I believe we could pass in the board id as either a hash or get parameter and then use JavaScript to check if that was passed in and just replace a few pieces of info on the unknown page on the fly. The main things I was thinking of replacing are:
1. The board title
2. The Browse S3 link under absolute newest
3. The Browse S3 link under Past Releases (which I think is always identical to the S3 link above)

This would at least give a semblance of usefulness to the unknown page until a board is added.

Neradoc:
resolves #1025 

- adds `?unknown_id=...` to the unknown boards URL
- the javascript on the board page looks for it
  - sets the title (and html title) by replacing underscores with spaces and uppercasing the first letter
  - replaces "unknown" with the ID in the URLs href, making them point to the right place

### DIFF
--- a/assets/javascript/download.js
+++ b/assets/javascript/download.js
@@ -19,6 +19,24 @@ document.addEventListener('DOMContentLoaded',function() {
     script.setAttribute('src', '//accounts.adafruit.com/users/locale?callback=setLocale');
     document.body.appendChild(script);
   }
+
+  // set the board ID based on query string
+  const hereurl = new URL(window.location.href);
+  const board_id = hereurl.searchParams.get("unknown_id");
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
Tested this out and it works great.

PR #1028, opened 2022-08-23
ELECFREAKS Pico:ed have a display but also Alligator Clip

dglaude:
Should we consider a 7x17 LED matrix to be a display?
I believe only LCD or OLED display should be tagged as having a display.
However I added the Alligator Clip friendly as it looks like a Micro:bit.

### DIFF
--- a/_board/elecfreaks_picoed.md
+++ b/_board/elecfreaks_picoed.md
@@ -10,7 +10,7 @@ date_added: 2022-4-21
 family: raspberrypi
 features:
   - Speaker
-  - Display
+  - Solder-Free Alligator Clip
 ---
 
 The **Pico:ed** is a development board based on **Raspberry Pi RP2040 MCU**. It uses dual-core Arm Cortex-M0+ processor with 264KB RAM. The front of the board contains two buttons and a 7x17 dot matrix screen, which can be conveniently used for classroom teaching.
--- a/_board/elecfreaks_picoed.md
+++ b/_board/elecfreaks_picoed.md
@@ -10,6 +10,7 @@ date_added: 2022-4-21
 family: raspberrypi
 features:
   - Speaker
+  - Display
   - Solder-Free Alligator Clip
 ---
 


dhalbert:
I think @deshipu would say that LED matrices are displays. They are low-resolution, but they can be used for gaming and scrolling text display.

dglaude:
One could say that the 5x5 from Micro:bit is a display... so it is hard to put a limit. My idea was that OLED or LCDs are screen, and a bunch of LEDs is not.

I did see the [pewpew](https://circuitpython.org/board/pewpew10/) from @deshipu and I did not react the same way as for Pico:ed or the two mixgo_ce_* so I am obviously biased, maybe the joystic presence made it clear to me that this was a "screen".

Maybe for display it is not such a big deal, if it is visible in the picture, and not the kind of display the user expect, it will be discarded. For other feature, if a board pretending to have Wifi, but wifi that is not supported by CircuitPython... then I would prefer to have that feature not tagged for that board (and it is the case for the only offender I know off, the Wio Terminal).

If we decide this board has a screen, then I still need to add the "Solder-Free Alligator Clip" but it can be another PR.

dglaude:
> The way I think about it is if somebody is filtering for boards with displays for some project and they saw a few with LEDs would they be okay with having that as an option?

Feel free to discard/close the PR not matching that or maybe let's put the question as "In the Weed" and wait for a decision there.

makermelissa:
I'd like to think about this some more before closing or merging.

makermelissa:
For the terms, let's go with:
Built-in Display
External Display Connector

"External Display Connector "could be added to the bare metal Raspberry Pi SBCs and similar boards as well.

dglaude:
Keeping the PR open for the Aligator Clip feature that was missing, but not removing the Display as discussion on 20220829 decided those LED matrix are small display.

makermelissa:
Actually, I'll use #1032.

makermelissa (approved):
Oh right, looks good.

PR #1029, opened 2022-08-23
Those two board do not really have a "display"

dglaude:
Except if we consider "16x8 blue LED dot matrix screen" to be a display... but then Micro:bit also have a display!


dglaude:
It's a display by 20220829 meeting.

PR #1030, opened 2022-08-23
Boards without a "Display" but with a "connector" for one.

dglaude:
This beetle-esp32-c3 board has a connector so that you can attach a display sold separatly.
Do we consider that a display?

From the documentation:
* "Besides, the onboard easy-to-connect GDI saves the trouble of wiring when using a screen."
* "Comes with expansion board, more convenient to make projects or use with a screen"

### DIFF
--- a/_board/beetle-esp32-c3.md
+++ b/_board/beetle-esp32-c3.md
@@ -15,7 +15,6 @@ features:
   - Battery Charging
   - Bluetooth/BTLE
   - Breadboard-Friendly
-  - Display
   - USB-C
   - Wi-Fi
 ---
--- a/_board/beetle-esp32-c3.md
+++ b/_board/beetle-esp32-c3.md
@@ -15,6 +15,7 @@ features:
   - Battery Charging
   - Bluetooth/BTLE
   - Breadboard-Friendly
+  - External Display
   - USB-C
   - Wi-Fi
 ---
--- a/_board/matrixportal_m4.md
+++ b/_board/matrixportal_m4.md
@@ -10,7 +10,7 @@ date_added: 2020-9-16
 family: atmel-samd
 bootloader_id: matrixportal_m4
 features:
-  - Display
+  - External Display
   - Wi-Fi
   - STEMMA QT/QWIIC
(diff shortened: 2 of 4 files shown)


dglaude:
Will have to find a proper name for the Feature: "Made for a Display".
Native English speaker might help here...this will apply to this board with a GDI connect and the two HUB75 discussed in #1032 
I don't know yet how to add a new feature (or I forgot) so once I figure that I will put all the change in this one PR.

makermelissa:
"Intended for External Display" perhaps

makermelissa:
It will need to be changed on all boards as well as the template. It's kind of tedious at the moment.

deshipu:
Maybe "Optional Display"?

kattni:
"Display ready"?

dglaude:
Ok, I'll start putting on the boards, currently that is only 3 boards...
For the template, I don't really know where that is.

Current options (need a voting place?):
"External Display Connector"
"Intended for External Display"
"Optional Display"
"Display Friendly"
"Display Ready"

makermelissa:
I updated issue #1032 to be about the name, so that would be a better place to keep the suggestions together.

dglaude:
> I updated issue #1032 to be about the name, so that would be a better place to keep the suggestions together.
So the result of a vote in Discord made "External Display" the winner.
I'll try to add that to the 3 boards identified, but if there are other change needed to make a new feature "real" I will need your help.

dglaude:
I think all the board that need tagging are tagged... but I don't know what else to modify.

I found the build.yml and I see where the call to `check-boards.py` is but I don't know what check are made and if there is something I can modify to make it pass the test:
```
    - name: Check feature names and other data
      run: python3 check-boards.py
```

makermelissa:
It will need to be updated in https://github.com/adafruit/circuitpython-org/blob/main/template.md as well. That's what the comparison is being made to for regular boards. For blinka boards, I just have values in the checks file until I think of something better.

makermelissa (approved):
Looks good.

PR #1038, opened 2022-08-29
There is no such board as pewpew13

deshipu:
I have no idea how it got there.


dhalbert:
It is mentioned in https://github.com/adafruit/circuitpython/blob/main/docs/shared_bindings_matrix.py

dhalbert:
@Neradoc `pewpew13` was added here: https://github.com/adafruit/circuitpython/pull/4539; maybe just a keyboard typo?

deshipu:
In any case, we don't need it.

makermelissa:
@deshipu, it looks like there's a merge conflict caused by #1037 that I don't have access to resolve.

Neradoc:
I added the name from the list of aliases back then, but the alias comes all the way from https://github.com/adafruit/circuitpython/pull/3722 it seems.

dhalbert:
The original reference appears to be even older. It looks it was originally `PewPew 10.3`, mentioned here: https://github.com/adafruit/circuitpython/issues/1673#issuecomment-474962752, and then `pewpew13` was added as an alias in #1781.

dhalbert (approved):

makermelissa:
Thanks @dhalbert for removing it in https://github.com/adafruit/circuitpython/pull/6845 as well so it doesn't come back.

PR #1040, opened 2022-09-03
Update adabot for latest RTD error specification

tekktrik:
Updates to a new commit of adabot to allow for more specific errors regarding the RTD documentation failure.  This will help with figuring out why adabot is having trouble, but no functional changes.


makermelissa (approved):
Looks good to me.

PR #1042, opened 2022-09-05
Lolin C3 is available in CP 8.0

dglaude:
8.0 is a major release (even if in Beta), so that part of the description does not make sense anymore.


makermelissa (approved):
Ah, thank you.

PR #1043, opened 2022-09-08
Update Seeed_XIAO_nRF52840_Sense.md

infamy:
Not sure what the best format would be.. but getting into the bootloader on this board is a bit odd, since it does not have a boot button... so figure it maybe worth placing it here.


Neradoc:
Hi, double pressing reset is the default instruction in the learn guide. It's common to every NRF and Atmel board, and all Adafruit boards (RP2040 and ESP32-Sx). There is no need for a special case for this board.

PR #1051, opened 2022-09-18
Added board files for TinyPICO and TinyPICO NANO

UnexpectedMaker:
These are my boards using the original ESP32-PICO-D4 chip form Espressif that are now supported in CP.


makermelissa (approved):
Thank you.

PR #1061, opened 2022-10-06
Original images are no longer deployed

### LINKED ISSUE #1057: Don't deploy "original" images
The pages deployment deploys all the versions of the board images: `original/`, `large/`, and `small/`, but I don't see that the `original/` images are linked anywhere. Some of those are very large, and cause the deployment artifact to be over 300MB. The deployment takes many minutes. Could we avoid deploying that directory?

makermelissa:
Fixes #1057.

### DIFF
--- a/_config.yml
+++ b/_config.yml
@@ -62,3 +62,4 @@ exclude:
   - template.md
   - Rakefile
   - .idea
+  - assets/images/boards/original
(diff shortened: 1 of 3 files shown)


dhalbert (approved):
Thank you for this!

dhalbert:
Build and deployment time went from 3 minutes (best case) to 1 minute. Artifact size is now <80MB, was about 325MB.

PR #1069, opened 2022-10-10
Make it clearer that broadcom builds are very alpha

dhalbert:
This adds more obvious warnings (and fixes grammar a bit) that the broadcom builds are very alpha and have significant issues. I've encountered this as a support issue several times recently. It's in the release notes but those are often not read.


makermelissa (approved):
Thanks, looks good.

PR #1070, opened 2022-10-11
Add BananaPi BPI-PicoW-S2 board.

Wind-stormger:
Adds support for the BananaPi BPI-PicoW-S3 Boards.
Depends on  https://github.com/adafruit/circuitpython/pull/7031


dhalbert (approved):
https://github.com/adafruit/circuitpython/pull/7031 has been merged, so adding here.

PR #1074, opened 2022-10-17
Added new Blinka Boards, fixed others

makermelissa:
This adds a couple of new boards that were merged in this morning and fixes a couple items on existing boards that I had learned more details about since they were added.


kattni (approved):
Looks good to me!

PR #1088, opened 2022-11-02
Added tags and module search feature

### LINKED ISSUE #1052: Feature request: search board by support matrix module
Here is the usecase scenario and interest of the feature (I have no clue how that could be implemented):

A user want to do something specific with CircuitPython and need to figure out if one of the board he has or he could want to acquire support his kind of hardware and has the specific needed module build-in.

So if I want to have a board that can be used to read from a PS2 keyboard, I can search in the support matrix for `ps2io` module like this: https://docs.circuitpython.org/en/latest/shared-bindings/support_matrix.html?filter=ps2io#module-support-matrix

It would be great to be able to search for a board directly in a filter, and visualize.
Of course it would need to get the information from the CircuitPython main repo, just like the support matrix.

Potentially some of the current manually maintained Feature could be translated into this new feature search.

Regards

### LINKED ISSUE #1064: Find the picow when searching for picow
Is there a way to have the Pico W show when searching "picow" (no space) ?
We could just add it to the name I guess, but isn't there a keyword field ?
And also "🥧🐮". No ? Ok not that one.

makermelissa:
Fixes #1064. Fixes #1052.

### DIFF
--- a/assets/javascript/downloads.js
+++ b/assets/javascript/downloads.js
@@ -421,8 +421,17 @@ function shouldDisplayDownload(download, displayedManufacturers, displayedMcufam
 
   if (downloadsSearch.searchTerm && downloadsSearch.searchTerm.length > 0 && shouldDisplay) {
     var regex = new RegExp(downloadsSearch.searchTerm, "gi");
-    var haystack = download.dataset.name + " " + download.dataset.id + " " + download.dataset.manufacturer + " " + download.dataset.mcufamily + " " + download.dataset.features;
+    var dataFields = [
+        download.dataset.name,
+        download.dataset.id,
+        download.dataset.manufacturer,
+        download.dataset.mcufamily,
+        download.dataset.features,
+        download.dataset.tags,
+        download.dataset.modules,
+    ];
 
+    var haystack = dataFields.join(" ");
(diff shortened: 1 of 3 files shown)


assets/javascript/downloads.js
```diff
@@ -421,8 +421,17 @@ function shouldDisplayDownload(download, displayedManufacturers, displayedMcufam
 
   if (downloadsSearch.searchTerm && downloadsSearch.searchTerm.length > 0 && shouldDisplay) {
     var regex = new RegExp(downloadsSearch.searchTerm, "gi");
-    var haystack = download.dataset.name + " " + download.dataset.id + " " + download.dataset.manufacturer + " " + download.dataset.mcufamily + " " + download.dataset.features;
-
+    var dataFields = [
+        download.dataset.name,
```
FoamyGuy, line 425: Thanks for changing this to a list. Much easier to read and add things to if anything else ever pops up.

FoamyGuy (approved):
Looks good to me. Tested successfully by running locally and trying the new search terms for picow. 

It's great to be able to fine tune the search results with tags for specific devices like this. 

Thanks for adding this functionality @makermelissa

PR #1090, opened 2022-11-06
Designate labels for checkbox in filters

jepler:
This means that the user can hit anywhere on the big target like "Adafruit" instead of only being able to click the small checkbox. It may also improve accessibility.

Additionally, the mouse cursor over the checkbox and its label is changed to the "pointer" to indicate an action is available by clicking.


makermelissa (approved):
Oh, good one. I didn't realize it wasn't working like this. Looks good. Didn't test.

PR #1092, opened 2022-11-07
Update to adabot with PyGithub 1.57

### LINKED ISSUE #1091: Adabot submodule needs to be fixed and repinned
The cron the updates the contributing information is failing because Adabot's current method of using validators is failing for Python 3.11.  I've submitted https://github.com/adafruit/adabot/pull/309 for this - hopefully that's the only change needed.  After that's merged, the `adabot` submodule should be updated here.

tekktrik:
Resolves #1091


makermelissa (approved):
Thanks

PR #1105, opened 2022-11-16
added/updated smart bee designs boards

strid3r21:
updated the bee motion s3 to add some addition information

added Bee s3 board relevant information/docs.


makermelissa:
Hi, looks like you're using a feature value that's different than our search filters. If it has battery charging, you can use "Battery Charging", otherwise it's probably best to remove it.

strid3r21:
> Hi, looks like you're using a feature value that's different than our search filters. If it has battery charging, you can use "Battery Charging", otherwise it's probably best to remove it.

sorry about that. it should be removed now.

makermelissa (approved):

PR #1113, opened 2022-12-09
Replace link for ESP Web Flasher

unwiredben:
The previous link went to a version that wasn't being maintained and had a banner directing users to a different tool. Replace with a link to the version of ESP Web Flasher hosted on adafruit.github.io.


makermelissa (approved):
Oh, good catch. I forgot about links on here.

PR #1114, opened 2022-12-12
Highlight incompatibility with Adafruit ecosystem

dglaude:
I try to manage expectation about this board and avoid short circuit or broken sensor. 1) I almost burned myself by mixing LOLIN I2C JST SH accessories and Adafruit QTPy, reverse way, I could have destroyed an Adafruit sensor by connecting it to Lolin S2 Pico. 2) The screen does not work as expected for a build-in screen, DISPLAY is not defined when you `import board`

I did hesitate between making a special section like "Caveats" or highlighting how it deviate from one can expect from a board running CircuitPython.


deshipu:
I suppose the display initialization could be added to board.c for this board?

dglaude:
Hi @deshipu , yes there must be a way to have the screen work as expected but apparently it is the first board where the screen is connected in I2C so there was no example to copy from.
https://github.com/adafruit/circuitpython/issues/7325#issuecomment-1345048637

If this PR make's it, and then later someone implement that the proper way, we can remove the warning (or say starting with CP9.0.0-alpha1 this statement is not true anymore).

evildave666:
To be fair, the S2 Pico circuitpython-org entry (I wrote it) does have a mention that the connector is not Stemma-QT pinout.

_board/lolin_s3.md
```diff
@@ -32,7 +32,7 @@ WiFi & Bluetooth 5 (LE) development boards based ESP32-S3-WROOM-1.
   - `EN` RESET button
   - `GPIO0` BOOT button
   - `GPIO38` WS2812B RGB LED 
-  - Lolin I2C JST SH 4-pin port (not QWIIC/Stemma-Qt pinout) using `GPIO42` (SDA) and `GPIO41` (SCL)
+  - Lolin I2C JST SH 4-pin port (**not QWIIC/Stemma-Qt pinout**) using `GPIO42` (SDA) and `GPIO41` (SCL)
```
dhalbert, line 35:
```suggestion
  - Lolin I2C JST SH 4-pin port (**does not match QWIIC/Stemma-Qt pinout**) using `GPIO42` (SDA) and `GPIO41` (SCL)
```

dhalbert (changes requested):
Thanks! One suggestion for increased clarity.

dglaude:
There might be a problem with the CI and Jekyll in particular, not sure why those simple change break it.

jepler (approved):
Approving based on Dan's previous review

PR #1122, opened 2022-12-29
Pin ruby at 3.1.

jepler:
.. it looks like the non-string value `3.0` was probably being interpreted as `3` aka `3.x`?

Previously it looks like we built with 'ruby 3.1.3p185 (2022-11-24 revision 1a6b16756e) [x86_64-linux]' so I chose to pin at "3.1"


dhalbert (approved):
The same thing "non-string bare integer value) happened with Python.

PR #1123, opened 2022-12-31
Update Lolin C3 Mini page to reflect new targetted hardware version

evildave666:
Update the Lolin C3 Mini page to reflect https://github.com/adafruit/circuitpython/pull/7392 changing the targeted hardware version to v2.1.

Status LED removed, replaced by WS2812B.
PCB antenna removed, replaced by ceramic antenna.

Is the language describing the RGB LED correct for a non-Adafruit (and presumably not licensed to use the term NeoPixel) WS2812B correct?


makermelissa (approved):
Looks good. Thanks.

PR #1128, opened 2023-01-11
Clarify which version of Luatos Core ESP32-C3 board is targeted

evildave666:
There are 2 versions of this board differing in what is connected to the USB-C port. This PR clarifies that the board definition is intended to target the newer version using the ESP32-C3 built-in USB-CDC/JTAG pins rather than via a CH343 UART to USB chip.


makermelissa (approved):
Awesome. Thanks.

PR #1132, opened 2023-01-18
Document the use of GPIO11 on the Luatos Core ESP32-C3 board

evildave666:
GPIO11 on the ESP32-C3 is by default dedicated as the VDD pin for external SPI flash.

This board does not use the pin for that purpose, instead directly connecting flash VDD to the 3.3V bus and breaks the GPIO out to a board pin. In order to actually use it as a GPIO in Circuitpython requires setting an efuse on the chip (and beyond that will require modifications to ports/espressif/common_hal/microcontroller/Pin.c) and building a custom image. I removed the GPIO from pins.c in https://github.com/adafruit/circuitpython/pull/7372

This PR documents that information in the notes section.


tannewt (approved):
Thank you!

PR #1133, opened 2023-01-25
Remove files pertaining to a board that will not be released

jepler:
See https://github.com/adafruit/circuitpython/pull/7478 -- should probably only be merged after that PR is merged.


makermelissa (approved):
Looks good. If the board will never be released, is there a reason it needs to wait for the PR?

dhalbert:
Other PR is now merged, so I merged this.

PR #1137, opened 2023-01-30
Initial CircuitPython Installer Coding

makermelissa:
I need to put this in place like this for now. Right now it will show the button, but when clicked it should show an empty menu with the message "Coming soon. Check back later.". Adding `#mode=test` to the URL and refreshing will show the different menu items. It's currently not completely functional which is why it's disabled by default, but this will allow me to test CORS from the real domain. Once I verify what needs to be done with CORS, I can make those changes and push subsequent updates to show the working install flows as I finish them.


tannewt (approved):
Happy to have this available for testing. One weirdness is the three files named `cpinstaller.js`, `installer.js` and `wsinstaller.js`. Could they be renamed to be clearer?

makermelissa:
Oh, `wsinstaller.js` shouldn't have been included, so I'll delete that. `installer.js` is the base installer (so it can eventually be reused for wippersnapper) which I can rename, and `cpinstaller.js` is the CircuitPython specific stuff. I'd like to have the installer eventually live in its own repo, so it will eventually go away from here and possibly the cpinstaller.js may move as well so it could be reused on other sites.

PR #1145, opened 2023-02-08
Change LiPoly chip from LC709203 to MAX17048.

miltieIV2:
LiPoly chip changed from LC709203 to MAX17048.
It also breaks this part of the guide: https://learn.adafruit.com/adafruit-esp32-s2-tft-feather?view=all#measuring-battery-3122383


makermelissa (approved):
Good catch. I had to check where you got the info from and noticed it was in fact updated on https://www.adafruit.com/product/5345, which is where I originally pulled this from.

PR #1146, opened 2023-02-14
Update actions to use Node.js 16

tekktrik:
Updates CI to use actions running on Node.js 16 to avoid deprecation


makermelissa (approved):
Looks good. Thanks.

PR #1149, opened 2023-02-18
CircuitPython Installer is now functioning

makermelissa:
Ok, this is pretty much functioning well at this point. I'm putting this in a draft because I want to polish it up a bit more and add some things like showing a link to the Code Editor if credentials have been added. I still haven't figured out a good way to get the IP address without logging into the REPL (for when install is done using the CIRCUITPY drive). I'll probably just show the circuitpython.local in that case.

There are some binaries included. Those are temporary because my computer throws CORS issues, so they will be removed eventually.


makermelissa:
This should be good to go now. It's not perfect, but it is in a good place to be added to the site.

makermelissa:
Found and fixed a bug while implementing a simple idea to make testing much easier.

cpinstaller/src/cpinstaller.js
```diff
@@ -391,466 +792,371 @@ export class CPInstallButton extends InstallButton {
...
-    }
-}
+    async getBootDisabled() {
+        // This is a very simple check for now. If there is something more complicated like a disable
+        // command behind an if statement, this will not detect it is enabled.
+        let fileContents;
+        if (this.repl) {
+            return true; // Always disabled in this case
+        } else if (this.circuitpyDriveHandle) {
+            fileContents = await this.readFile("boot.py");
```
tannewt, line 947: Can you hit this case? I don't think you can have a handle to the CIRCUITPY drive if it is disabled.

cpinstaller/src/cpinstaller.js
```diff
@@ -391,466 +792,371 @@ export class CPInstallButton extends InstallButton {
...
+    saveBootDisabled(disabled) {
+        // TODO: Save/remove a copy of boot.py on the CIRCUITPY drive
+        // This depends on whether it is currently disabled in boot.py and what the value of disabled is
+        // If they are the same, we can skip
+        // An idea is to only deal with this if boot.py doesn't exist and just use a generic boot.py
+        // For disabling, we can compare to the generic and if they are different refuse to touch it
+        const formElement = this.currentDialogElement.querySelector('#circuitpy_drive');
+        if (formElement) {
+            if (formElement.checked) {
+                this.tomlSettings['CIRCUITPY_DRIVE'] = "disabled";
```
tannewt, line 971: Should we add support to CP for this?

cpinstaller/src/boot.py
```diff
@@ -0,0 +1,3 @@
+import storage
+
+storage.disable_usb_drive()
```
tannewt, line 3: Is this ever copied over?

tannewt (changes requested):
A couple questions around boot.py. Otherwise this looks good!

cpinstaller/src/cpinstaller.js
makermelissa: When the UF2 is first installed, the CIRCUITPY drive is enabled, which makes writing through the REPL impossible. So for boards with the CIRCUITPY drive it currently needs to take this route until it is able to be disabled (a future feature).

cpinstaller/src/cpinstaller.js
makermelissa: Actually, that would make things much easier from the installer if we could. That was just some temporary stuff that's not currently being called.

cpinstaller/src/boot.py
makermelissa: Not yet. I was trying to get something up for now since trying to disable the boot.py was ending up being more complicated than I thought, but that was planned to be part of another update.

makermelissa:
Basically, the boot.py stuff is not currently being utilized because it ended up being a much more complicated feature than I had originally thought and the project was stalled as a result. I was planning on following up in another PR.

cpinstaller/src/cpinstaller.js
makermelissa: One of the reasons for this is maybe somebody wants to use the installer, but doesn't want to use the Web Workflow to edit files and feels more comfortable using the CIRCUITPY drive.

I should also add, one of my goals was to keep user interaction as minimal as possible, so if they are doing web workflow, I was thinking of disabling CIRCUITPY at the last moment. But upon further thinking, it may be better to do that right away and then use the REPL so that we can get the IP address. For now, I'm deferring the decision.

tannewt (approved):
Ok, sounds good! I'm excited for people to start trying this. We can sort out the drive later.

tannewt:
(merge when you are ready)

PR #1151, opened 2023-02-21
Upated the Brainboardz_Neuron .md File to be Compliant with the Correct Features Terminology

BrainBoardz:
Hi There: 

I have completed the .md file for brainboardz_neuron in the _board directory and I have added the (original, large and small images) for the brainboardz_neuron to the assets/images/boards/ folders. I have also updated the Features section of the .md file to be compliant with the instructions on this item. 

If this can get merged and become active I would greatly appreciate it. Thanks for all the help!


FoamyGuy (changes requested):
Thank you. 

I tested this out locally and it looks mostly good. One tweak to the purchase header is the only improvement I could find.

_board/brainboardz_neuron.md
```diff
@@ -1,16 +1,48 @@
...
+ - 14 Touch Sensors 
+ - Blink LED
+ - MicroSD Card Reader
+ - Micro-USB Connector
+ - Low Noise TI 5V to 3.3V LDO
+ - 2.54mm 4 Pin Header4 (Neuron-Z) Boards
+ - Access to All of the ESP32-S3's GPIO Pins
+ - Compatible with all BrainBoardz boards
+ 
+ ## Purchase:
```
FoamyGuy, line 46: There is an extra space at the beginning of this line that is causing the header syntax to get mis-interpreted:
image

With this change it should render as a header instead:
```suggestion
## Purchase:
```

FoamyGuy (changes requested):
Thank you. 

I tested this out locally and it looks mostly good. One tweak to the purchase header is the only improvement I could find.

BrainBoardz:
Hi foamguy:

 I made the change as requested and committed it to my branch. I'm a bit
unsure whether I need to submit another pull request. Github is new
territory for me.
If I commit to the main branch is that ok? I could do another pull request
but I think that is not how changes are supposed to work. Sorry for
the confusion.

FoamyGuy (approved):
@BrainBoardz once you have created the PR you can add new commits to the same branch (`BrainBoardz:main` in your case) and the PR will automatically update to reflect the new commits. The PR tracks the branch if/when it changes, so there is no need to make a new PR, making new commit(s) is good enough.

The latest version looks good to me. Thank you!

BrainBoardz:
Hi  There:

That's awesome news. Not only did it get posted (yay!) but I now understand
how the process is supposed to work. Thanks for all the help!

bye for now,

BrainBoardz:
Great. Thanks!

Issue #1121, opened 2022-12-29
jekyll vs ruby 3.2.0 incompatibility


jepler:
See https://github.com/jekyll/jekyll/issues/9231

investigating how to install an older version of ruby pending a fix at jekyll and/or liquid.

makermelissa:
It looks like this has been fixed as of 2 days ago by using liquid 4.0.4 (see https://github.com/jekyll/jekyll/issues/9231#issuecomment-1379146408), so I think unpinning is all that is needed.

makermelissa:
That and running bundle update to update the Gemfile.lock.

jepler:
protobuf also isn't able to support ruby 3.2.0 yet: https://github.com/protocolbuffers/protobuf/issues/11538

makermelissa:
Ah, ok.

jepler:
it comes in as an indirect dependency related to sass, I don't understand why

PR #1156, opened 2023-02-26
bump ruby & gem versions

jepler:
~~Let's see if #1121 is fixed upstream yet.~~

Closes: #1121


makermelissa (approved):
Looks good to me.

PR #1160, opened 2023-03-01
Update Installer to use selected language

makermelissa:
This also sets it to use Amazon S3 for the download link for now to get around some CORS problems that cloudfront is causing and the "CircuitPython Installer" title is moved out of the base installer and into the CP specific one.

Also, uncommented some steps for the Bin file install so that it erases the chip first.

### DIFF
--- a/cpinstaller/src/cpinstaller.js
+++ b/cpinstaller/src/cpinstaller.js
@@ -28,15 +28,21 @@ const FAMILY_TO_CHIP_MAP = {
     'esp32': esptoolPackage.CHIP_FAMILY_ESP32
 }
 
+const attrMap = {
+    "bootloader": "bootloaderUrl",
+    "uf2file": "uf2FileUrl",
+    "binfile": "binFileUrl"
+}
+
 export class CPInstallButton extends InstallButton {
     constructor() {
         super();
         this.releaseVersion = "[version]";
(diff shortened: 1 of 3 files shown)


cpinstaller/src/cpinstaller.js
```diff
@@ -55,32 +61,36 @@ export class CPInstallButton extends InstallButton {
...
+        this[classVar] = currentValue ? this.updateBinaryUrl(currentValue) : null;
+    }
+
+    updateBinaryUrl(url) {
+        //if (location.hostname == "localhost") {
             if (url) {
                 url = url.replace("https://downloads.circuitpython.org/", "https://adafruit-circuit-python.s3.amazonaws.com/");
             }
-        }
+        //}
```
dhalbert, line 93: Just checking that this is not leftover debugging commented-out stuff

dhalbert (commented):
Thanks - one q about commented-out code

cpinstaller/src/cpinstaller.js
makermelissa: Correct. That's just so it will use Amazon S3 for the download on everything. Once we figure out why cloudfront keeps refusing to give the correct allow headers, we can revert it back by uncommenting.

dhalbert (approved):

PR #1165, opened 2023-03-13
Board Info Generator Script

makermelissa:
As part of getting the CircuitPython Installer ready for learn, it's useful to have the necessary board info in a JSON file so that just the board id (or IDs) is all that will need to be provided. This PR adds a script to generate and upload the file to Amazon S3.

I plan on having these values overridable with the button parameters so that multiple languages can still be installed. After this is merged, changes to the installer to grab the info from this file will be required, which will be part of a follow-up PR.


makermelissa:
For some reason, the file is not actually uploading, so this is still being worked on.

makermelissa:
I was getting a bit perplexed by this and then I realized it probably won't actually upload a file to Amazon S3 until after this has been merged for security reasons. So this should be good to go.

dhalbert (changes requested):
Looks fine, but there is no newline at the end of `generate-board-info.py`. This is hard to fix from the GitHub web interface. Can you push a commit? Thanks.

makermelissa:
I just ran black so it would fix any other formatting issues.

dhalbert (approved):

PR #1168, opened 2023-03-14
Update site to point to externally published cpinstaller

makermelissa:
This updates the site to point to the installer published at https://github.com/adafruit/web-firmware-installer-js. It also removes some parameters that are no longer needed because they are extracted from the esp32_boards.json file.


makermelissa:
This should be all ready to go. I've tested a lot and made a few small changes along the way to fix any issues I've found, so it's more robust than what is currently in place.

dhalbert (approved):
Excellent!

PR #1175, opened 2023-03-29
Change Naming Style of Deneyap Boards

DogushC:
Fix the problem about not listing .bin or .uf2 file when clicking "Browse S3" button on own page of the board because of naming style of Deneyap boards. 

Delete contribute section.


dhalbert (approved):
:+1:

makermelissa:
GitHub Actions is currently down, so this will not automatically be updated. The deploy action will need to be run manually later.

PR #1177, opened 2023-03-30
Update Board IDs Deneyap Kart/1A/G

DogushC:
Update Board IDs of Deneyap Kart, Deneyap Kart 1A and Deneyap Kart G.


makermelissa (approved):
Thanks, looks good.

PR #1181, opened 2023-04-04
Add Feather RP2040 RFM boards.

kattni:
As per discussion on Discord, one is aliased to the other as it was the only way to get them both to show up. Everything passed and built locally.


makermelissa (approved):
Looks good to me. Thanks.

PR #1186, opened 2023-04-09
Fig pi Update

bwshockley:
Updated the board description and links.

Added newer images in the original, small, and large sizes.


tannewt (approved):
Looks good!

PR #1190, opened 2023-05-02
Adding USB Host and ThinkInk Feathers

BlitzCityDIY:
Adding the USB Host and ThinkInk Feathers to _boards


jepler (approved):
Looks good! Not sure why so many images needed to be compressed but that's fine

jwcooper:
@jepler i think the image compression action happens when a pr is not from a fork. The action won't run on forks, so they get queued up. (I think)

PR #1192, opened 2023-05-02
Add process to Flash Circuitpython to the board

lboue:


RetiredWizard:
You can get there after a few clicks from the "MGM240P Wireless Module" link but should there be a direct link to the "Simplicity Commander" download as well?

lboue:
Hello @RetiredWizard:

1. Can I have confirmation of the addresses to use to flash the binary?
2. I propose to add some notes to help users:

## Method 2 - Simplicity Commander
You can use Simplicity Commander: A tool that contains a scriptable command line, allows the user to flash their application.

1. Download [Simplicity Commander](https://www.silabs.com/developers/mcu-programming-options)
2. Extract Simplicity Commander directory from the zip file.
3. Go to Simplicity Commander directory
4. Run commander.exe in commande line

Does this look correct?

makermelissa:
It sounds like this is still a work in progress, so I'm converting to a draft for now.

PR #1193, opened 2023-05-04
Pimoroni dv base pico

RetiredWizard:


makermelissa (changes requested):
Thanks for submitting. Could you please add the Adafruit product link as well? https://www.adafruit.com/product/5674. Also, please use an image without text if possible. You could use the one from the Adafruit Product link.

RetiredWizard:
Sure thing, I won't be home until tomorrow afternoon but I'll update it then. Thanks!

makermelissa (approved):
Thank you. Looks great.

PR #1195, opened 2023-05-06
Fix fullwidth parentheses in YD-RP2040 board description

shuyuan-liu:
This is a small formatting change.

The current description contains full width parentheses `（）`. They aren't intended for English text and were probably left in when translating from Chinese. These are changed to half width `()`.

Additionally two periods are removed from the list of improvements for consistency within the list.


makermelissa (approved):
Thank you. Looks good.

PR #1196, opened 2023-05-08
Update missing boards Batch 2

makermelissa:
Here are the rest of the Blinka boards. This is part 2. Part 1 was in #1194. This brings the count up to 120.


tannewt (commented):
Where did you get these images? What license are they under?

Want to include distro info on the pages?

makermelissa:
> Where did you get these images?
Various places. Let's See:
I got the LubanCat ones from Taobao
I got the ROC-RK3328-CC from Libre Computer's product page (which I linked to for purchase)
I got the Jetson Images from the Purchase Pages I linked to.
I got the Radxa ones from the Purchase pages I linked to.

> What license are they under?
I'm not sure. Finding images for these boards was difficult itself.

> Want to include distro info on the pages?
I think we should add that in another PR. I'm still thinking about how I want to format the info. I'm thinking I may want to use a list, but I want to probably add it to the checker scripts so we don't end up with 20 different spellings of Debian or maybe I'll just make it part of the description.

makermelissa:
> > I'm not sure. Finding images for these boards was difficult itself.
> 
> I'm wary about using pictures we don't necessarily have the rights to use. I realize that other images may be in the same bucket though.

Fair enough. How would you like to proceed?

tannewt (approved):
Discussed during our meeting and decided product photos are ok as long as we link to the shop where it came. (And we don't hear otherwise.)

PR #1200, opened 2023-05-15
Fix download links for RP2040 Feather ThinkInk

### LINKED ISSUE #1199: Downloading ThinkInk UF2 results in NoSuchKey error
From the ThinkInk page at https://circuitpython.org/board/adafruit_feather_rp2040_thinkink/ if you click the `Download .UF2 Now` button an error page pops up that says:

```
NoSuchKeyThe specified key does not exist.bin/adafruit_feather_rp2040_thinkink/en_US/adafruit-circuitpython-adafruit_feather_rp2040_thinkink-en_US-8.1.0-beta.2.uf2Q10MMZEDH3FGDR91oJjfGeJ70KCpvG+NzmmXEvh1OazNg1JlhRijglVqWem8ZZcRfGu3tYQReYewKMt94qg1llCMqTU=
```

makermelissa:
Fixes #1199. Apparently in the previous releases it was called `adafruit_feather_rp2040_epd`, but in future releases will be called `adafruit_feather_rp2040_thinkink` because of https://github.com/adafruit/circuitpython/pull/7927. I had to change it to `adafruit_feather_rp2040_epd` for now, but it will need to be reverted on the next release.

### DIFF
--- a/_board/adafruit_feather_rp2040_epd.md
+++ b/_board/adafruit_feather_rp2040_epd.md
@@ -1,7 +1,6 @@
 ---
 layout: download
-board_id: "adafruit_feather_rp2040_thinkink"
-board_alias: "adafruit_feather_rp2040_epd"
+board_id: "adafruit_feather_rp2040_epd"
 title: "Feather RP2040 ThinkInk Download"
 name: "Feather RP2040 ThinkInk"
 manufacturer: "Adafruit"
--- a/_board/adafruit_feather_rp2040_thinkink.md
+++ b/_board/adafruit_feather_rp2040_thinkink.md
@@ -1,6 +1,7 @@
 ---
 layout: download
 board_id: "adafruit_feather_rp2040_thinkink"
+board_alias: "adafruit_feather_rp2040_epd"
 title: "Feather RP2040 ThinkInk Download"
 name: "Feather RP2040 ThinkInk"
 manufacturer: "Adafruit"
--- a/_data/files.json
+++ b/_data/files.json
@@ -3902,7 +3902,7 @@
  },
  {
   "downloads": 0,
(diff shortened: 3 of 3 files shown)


tannewt:
Could we make a system to allow for name changes? Dan and I were talking about changing some older board names to include manufacturer. It'd be cool to redirect from the old name to the new and manages the download link based on version.

tannewt (approved):
Good for now. Thanks!

makermelissa:
> Could we make a system to allow for name changes? Dan and I were talking about changing some older board names to include manufacturer. It'd be cool to redirect from the old name to the new and manages the download link based on version.

Not easily, but probably. The Liquid language that Jekyll uses generates static pages based on the info in the .md files. The pages that appear to be dynamic such as the unknown page and the installer are done purely with JavaScript. To have it deal with the updates would require rewriting the Liquid code that generates the static pages so that it more heavily relies on the data files, which may not be such a bad idea really. Currently the code is very patchy, making introducing bugs a little too easy. I would like to take some time to clean it up. It would mostly just take some time because Liquid is not intuitive. I can create an issue for now.

tannewt:
We could consider abandoning Jekyll and Liquid all together. I think the static site generation options got much more flexible with GitHub actions.

makermelissa:
> We could consider abandoning Jekyll and Liquid all together. I think the static site generation options got much more flexible with GitHub actions.

Hmm, I'd be ok with that. I'll need to take a look at the other options to see what would might meet our needs better.

jepler:
There does appear to be a plugin for adding redirects in a simple way. no idea if it works. https://github.com/jekyll/jekyll-redirect-from

makermelissa:
Good to know for updating the epd board back to thinkink, but it could cause things to get complicated if we overuse it.

Issue #1110, opened 2022-11-29
Blinka page should not link CP firmware


dglaude:
On the Blinka side ( https://circuitpython.org/blinka ), we should not link to CircuitPython firmware like we do in https://circuitpython.org/downloads .

This is visible in the RP2040 board here: https://circuitpython.org/blinka?sort-by=downloads&manufacturers=Adafruit

Maybe do like for the Pico ( https://circuitpython.org/blinka/raspberry_pi_pico/ ) and link to https://learn.adafruit.com/circuitpython-libraries-on-micropython-using-the-raspberry-pi-pico (or similar learn guide).

PR #1201, opened 2023-05-15
Blinka Fixes

makermelissa:
Fixes the name of the Jetson Orin Nano and removes a board that was accidentally added to Blinka (0xCB Helios). It also checks that the blinka flag is set to True for Blinka boards to help mitigate this from happening again. The script did find one board missing the flag, so it was added.


makermelissa:
Fixes #1110 by adding a check to ensure the blinka filed is true, which makes it so the CP downloads don't show.

PR #1203, opened 2023-05-17
Update challenger_rp2040_lte.md

rcarteraz:
Minor spelling/grammar issue.


makermelissa (approved):
Thanks. A lot of the text for the boards comes from copy/pasting on the internet and fixing obvious things, so contributions like this are very helpful.

PR #1206, opened 2023-05-23
M5Stack Timer Camera X

bill88t:
It's 4 am almost. Enjoy.


bill88t:
Added `Camera` to features. It's such a basic thing. I don't get how this wasn't already one.

makermelissa (changes requested):
Please don't try and circumvent the checks by adding to the features list without previous agreement. The reason we don't have it on the list is because there are so few boards with this feature and we're trying to keep the features filters from being too long:
Screen Shot 2023-05-23 at 4 55 22 PM

bill88t:
Alright, the change has been reverted.

Note: Tag `GPS` only contains one board, while `Camera` (with this one) would have 4. And m5stack has quite a few more esp32 boards with cameras that may be added in the future.

makermelissa:
> Alright, the change has been reverted.
> 
> Note: Tag `GPS` only contains one board, while `Camera` (with this one) would have 4. And m5stack has quite a few more esp32 boards with cameras that may be added in the future.

Hmm, that's a good point. Then if you'd like to add Camera back in, but also add it to the other boards, I'd be ok with that. Sorry to go back and forth on that.

bill88t:
I have reintroduced the camera tag.

Would you perhaps want me to remove tag GPS to keep the list length the same?

makermelissa:
Let's leave the GPS filter for now.

PR #1211, opened 2023-05-24
add Lolin_C3_Pico board

tkroo:
add info for lolin c3 pico board


makermelissa (approved):
Thank you. Looks good.

PR #1213, opened 2023-06-05
Added missing boards from 8.1.0

makermelissa:


tannewt (changes requested):
Thank you for adding these! The 1040 picture is of the 1015. Please update it. Thanks!

makermelissa:
Good catch. Fixed.

tannewt (approved):
Thank you!

PR #1220, opened 2023-06-30
refer to guide pages for nRF bootloader updates

dhalbert:
The nRF bootloader update blurb referred to the README in the bootloader repo. There are now guide pages for this that are more comprehensive, so point to them.

### DIFF
--- a/_includes/download/board.html
+++ b/_includes/download/board.html
@@ -287,9 +287,9 @@ By the way, boolean operation precedence is right to left! (yeesh)
 
   {% if page.family == 'nrf52840' %}
   <p>
-    Updating the bootloader on nRF52840 boards is an involved process right now.
-    Follow the instructions in this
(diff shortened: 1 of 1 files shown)


dhalbert:
tagging a bunch of you for review and notice: just one review is fine

jepler:
Is it worth providing a note here like the one you'll put in the release notes? I don't know if you have written that language yet but something like
> **TinyUF2 bootloaders below version 0.6.1 cannot load CircuitPython 8.2.0. Many boards are factory programmed with an older TinyUF2 version. In bootloader mode, check your TinyUF2 version by looking in the file INFO_UF2.TXT and upgrade if necessary.**

dhalbert:
> Is it worth providing a note here like the one you'll put in the release notes?

Good idea. Pushed a revision. It's not literally TinyUF2, it's the original nRF52 bootloader.

jepler (approved):
thank you! looks good to me now.

PR #1223, opened 2023-07-11
Adding QT Py ESP32-S3 with PSRAM

BlitzCityDIY:
Adding the new QT Py ESP32-S3 with PSRAM


makermelissa (approved):

PR #1237, opened 2023-08-04
include more prominent warning to update bootloader on nRF boards

dhalbert:
Put a warning to to update the bootloader on nRF boards right in the download box for CircuitPython, not just in the bootloader box.

Not everyone reads the release notes before updating. This will save support time.


jepler (approved):
Thank you, should be a good improvement.

PR #1241, opened 2023-08-30
Add new board - waveshare esp32 s3 pico

tirtharajsinha:
Requesting to add waveshare esp32 s3 pico board.


tirtharajsinha (commented):
Updating board details.

PR #1243, opened 2023-08-30
Add New Board - Waveshare ESP32-S3-Pico

tirtharajsinha:
Requesting to add Waveshare ESP32-S3-Pico.


makermelissa (approved):
Looks good. Thank you.

PR #1244, opened 2023-08-31
Pimoroni Pico DV Base with Pico W

RetiredWizard:
I'm not sure the "Note" and example code makes sense for circuitpython.org but I figured I'd throw it in and if it doesn't make sense I can pull it out easily enough. Thanks!


dglaude:
Hi, not sure I have ever seen a piece of code on circuitpython.org.

Maybe the best place could be to create a simpletest_code.py next to the board definition in circuitpython repo, as suggested in the first in the weed topic of this week meeting: 
https://github.com/adafruit/adafruit-circuitpython-weekly-meeting/blob/main/2023/2023-08-28.md

Now that would be the first time it is ever done, there might be discussion on the best place to put that.

I would also insist on putting a bit of code to read the 3 buttons to demonstrate how to use that part of the board and kickstart anyone wanting to test the board with a new firmware without having to write any code.

Regards

RetiredWizard:
When I was posting this PR, I didn't see the code snippet as example code, I was thinking it was part of a board caution or warning. I had a feeling that some other boards did something similar when there was a minor revision that swapped a pin or two out or for the boards that don't have native USB. 

Now that I've thought about it more, and poked around on circuitpython.org, I think those notes/cautions were probably added to the Adafruit product pages. I'm thinking for community provided board definitions the board description on circuitpython.org is probably still the right place if some sort of board warning is needed.

That being said, I clearly added some example code which should conform with whatever standard is decided upon for example code designed to familiarize users with a board. I'll go ahead and remove the code snippet for now and wait to see how to handle sample code.

makermelissa (approved):
Thanks for adding. Looks good.

PR #1245, opened 2023-09-05
Renaming waveshare_esp32s3_pico.md to waveshare_esp32_s3_pico.

tirtharajsinha:
Updating board name because the name here does not match in circuitpython repo.


makermelissa (approved):
Looks good. Thank you.

PR #1246, opened 2023-09-06
New boards plus updated adafruit board images

makermelissa:
This adds 2 new Blinka Boards, 3 new CircuitPython boards, and updates 7 board images.


tannewt (approved):
Thanks for the update!

PR #1255, opened 2023-09-21
Don't promise BLE for Pico W

dhalbert:
BLE support on Pico W is not currently actively being worked on. Remove "coming soon" language and remove BLE as a feature. See https://github.com/adafruit/circuitpython/issues/7693.


makermelissa (approved):
Looks good to me.

PR #1266, opened 2023-10-21
Added image and site.mds for SuperMini NRF52840 and Waveshare ESP32-S3-Zero

Axeia:
I recently added support for these boards, this should add all the info so they can start showing up on the circuitpython.org website.


makermelissa (approved):
Thanks

PR #1277, opened 2023-11-14
Update Seeed_XIAO_nRF52840_Sense.md

dglaude:
I guess the vendor documentation was corrected as this is the new version of that sensor on https://www.seeedstudio.com/Seeed-XIAO-BLE-Sense-nRF52840-p-5253.html

I was super confuse to read that this board could also `speak through the the microphone`. I guess that was lost in translation.


makermelissa (approved):
Thanks. Yeah, I think the original sentence was a bit unclear even to a native English speaker.

PR #1278, opened 2023-11-14
Title and name entries indicate S3 not C6: espressif_esp32c6_devkitm_1_n4.md

rowsail:
Title and name entries indicate S3 not C6.


makermelissa (approved):
Thanks. Yeah, based on the filename, this change looks correct.

PR #1279, opened 2023-11-15
Update adafruit_metro_rp2040.md

eightycc:
With CircuitPython loaded on this bad boy you get about 15M of flash filesystem space.


dhalbert (approved):
:+1:

PR #1281, opened 2023-11-15
Add missing boards for CP 9.0.0-alpha.4 and CP 9.0.0-alpha.5

makermelissa:
Plus a new Blinka board from this morning...


dhalbert (changes requested):
The Elite-Pi picture is the bottom of the board. Maybe use the top? https://keeb.io/products/elite-pi-usb-c-pro-micro-replacement-rp2040

makermelissa:
Ok, changed.

dhalbert (approved):
Thanks for the new image!

PR #1284, opened 2023-11-21
Add Firebeetle 2 ESP32-S3

bill88t:
Please do take note of the added markdown. I can't test locally.

Mentioning `AXP313A` is pretty important as users will be unable to use the camera without it.

Also, even though the board has a display interface, I did not add the tag `Display`, since it doesn't ship with one.
Meanwhile, for the camera I did add the tag since it ships with a OV2640.


makermelissa (approved):
Thanks for adding. The additional markdown is fine.

PR #1285, opened 2023-11-22
Added TinyC6 and updated a bunch of other images for my boards

UnexpectedMaker:
Added my new TinyC6 to the boards and updated a bunch of board images to their latest versions.


makermelissa (approved):
Thank you for updating. You may want to also update the Bling, Blizzard S3, and TinyWATCH S3 descriptions. I wasn't able to find much info on those.

PR #1289, opened 2023-11-28
Remove adafruit esp32s2 camera

makermelissa:
The `adafruit_esp32s2_camera` has been revised to `adafruit_esp32s3_camera`. However, it still exists in the 8.2.x branch of CircuitPython, so it's possible that it may pop up again if a new 8.x release is made. This will hide it from the search results and if somebody goes to the URL, it will redirect them to the adafruit_esp32s2_camera page. Once CP 9.0 is the stable release, it can be removed.


dhalbert:
I can remove it from 8.2.x. We have several pending changes to 8.2.8, so there will be an 8.2.9 in the not too distant future.

makermelissa:
Ok, let's do that and I'll revise this to remove it.

dhalbert (approved):

PR #1294, opened 2023-12-01
pico_w: update the documentation link

DeadSix27:
The old one is gone so we link to the overall pico(w) documentation page


makermelissa (approved):
Thank you.

PR #1300, opened 2023-12-08
Fix board API output

### LINKED ISSUE #1299: Bad output from Board API
It appears the board api (https://circuitpython.org/api/boards.json) is just taking the boards page html output and putting all of the tags in a json file.

makermelissa:
Fixes #1299. Also, fixed one of the hidden boards' board_url format.


dhalbert (approved):
:+1:

PR #1301, opened 2023-12-11
Add missing boards for CP 8.2.9 plus more Adafruit links

makermelissa:


makermelissa:
I generally try and change text like that, so I just updated it.

dhalbert (changes requested):
wisdpi_tiny_rp2040.md needs a little copyediting to remove first-person references. It's also quite redundant: for instance the additional WS2812 RGB LED is mentioned three times (!).

I started to propose suggestions, but it got out of hand :)

makermelissa:
Yeah, I had copied and pasted from their website. I just revised most of it and updated.

dhalbert (approved):
Thanks for cleanining up the copy.

dhalbert:
The multiple links for boards is working out great! Thanks for adding that.

makermelissa:
You're welcome. :)

PR #1305, opened 2023-12-15
Image Compression Action Change

jwcooper:
Our current image compression doesn't run well due to the workflow of limiting to forked pull requests.

This proposed change modifies the image action to run once a month on the first of the month.

It will create a PR with compressed/optimized images that we can review and merge.

This is a suggested workflow from the image-action:
https://github.com/marketplace/actions/image-actions#compress-on-demand-or-on-schedule


makermelissa (approved):
Looks good to me.

PR #1310, opened 2024-01-02
Remove dev version of google-protobuf from Gemlock

makermelissa:
This appears to mostly be a duplicate, but a dev version. I think this will make it so actions stops failing.


jepler (approved):

jepler:
there's an incompatibility between jekyll and logger 1.4.3: https://github.com/jekyll/jekyll/pull/9392

PR #1311, opened 2024-01-02
Update jekyll to fix an incompatibility with logger

jepler:
I think this may fix the error during actions:
```
/opt/hostedtoolcache/Ruby/3.3.0/x64/lib/ruby/3.3.0/logger.rb:384:in `level': undefined method `[]' for nil (NoMethodError)

    @level_override[Fiber.current] || @level
                   ^^^^^^^^^^^^^^^
	from /home/runner/work/circuitpython-org/circuitpython-org/vendor/bundle/ruby/3.3.0/gems/jekyll-4.3.2/lib/jekyll/log_adapter.rb:45:in `adjust_verbosity'
```

### DIFF
--- a/Gemfile.lock
+++ b/Gemfile.lock
@@ -19,7 +19,7 @@ GEM
     image_processing (1.12.2)
       mini_magick (>= 4.9.5, < 5)
       ruby-vips (>= 2.0.17, < 3)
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Thanks

PR #1315, opened 2024-01-24
added Lichee Pi 4A

anonymouscowhead:
Add Sipeed Lichee Pi 4A SBC, recently added to PlatformDetect and Blinka


makermelissa (approved):
Thank you! I think you're the first person besides myself to add a Blinka board here.

PR #1316, opened 2024-01-25
Add PicoMo V2 board

supcik:
Here is the description for the PicoMo board developed at the School of Engineering and Architecture of Fribourg.


makermelissa (approved):
Thank you.

PR #1322, opened 2024-02-08
Improve search

### LINKED ISSUE #1321: board doesn't show up on circuitpython.org/downloads if searched for: e.g. feather tft should find feather s3 tft
### CircuitPython version

```python
latest
```

### Code/REPL

```python
n/a
```

### Behavior

search for `tft`, notice the 4 feathers amongst 5 results...
search for `feather tft`, notice only two reverse tft feathers.

I wanted the feather s3 tft at the time, but I'm sure the search issues relate to more boards.

### Description

_No response_

### Additional information

_No response_

jepler:
This turns the search function into more of a 'fuzzy search': The search is expected to be a series of space-separated terms, but the terms can appear in any order, can be just part of a word (e.g., "uefr" would match "bluefruit") and with other terms between them.

So for instance after this change both "tft feather" or "feather tft" will show the Feather ESP32-S3 TFT PSRAM (as well as 3 other feather boards)

Closes #1321

### DIFF
--- a/assets/javascript/downloads.js
+++ b/assets/javascript/downloads.js
@@ -387,6 +387,17 @@ function setFeaturesChecked() {
   downloadsSearch.featuresChecked = document.querySelectorAll('input[name="feature"]:checked').length > 0;
 }
 
+var cachedSearchTerm;
+var searchRx;
+function searchTermRegex() {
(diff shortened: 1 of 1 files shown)


makermelissa (approved):
Looks good to me. Much simpler than I was thinking it would be.

PR #1328, opened 2024-02-18
Refactor of search function from regex to string based, massive speedup in searchbar performance 

michalpokusa:
Current searching and filtering login implementation uses regexes, which leads to performance issues. If someone would type quickly "matrixportal", the search bar would hang on e.g. 8th character for a moment and then continue to render.

Additionally, current regex makes it that "bit matrix" and "matrix bit" produce different result, which i believe is not intended. 

Overall, this change speed up filtering (in my case, on my computer) from 400-700ms to 2-5ms and produces same results no matter the order of terms in search field.

🏗️ Refactor:

- Function `shouldDisplayDownload` that determined whether to show or not a card with board now uses string based logic instead of regex

Easy way to test this is simply copy-paste the function to console in browser and the search bar will use it from that moment until site reload.

### DIFF
--- a/assets/javascript/downloads.js
+++ b/assets/javascript/downloads.js
@@ -418,7 +418,7 @@ function shouldDisplayDownload(download, displayedManufacturers, displayedMcufam
     ].join(" ").toLowerCase();
 
 
-    for (const term of downloadsSearch.searchTerm.split(" ")) {
+    for (const term of downloadsSearch.searchTerm.toLowerCase().split(" ")) {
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Thanks for this! I tested with the old and new search in a number of cases and saw no differences.

jepler:
Thanks from me as well! the code I wrote was quick & dirty, and I hadn't noticed the performance problems.

PR #1333, opened 2024-02-21
add senseBox MCU-S2 ESP32S2

mariopesch:
add new senseBox MCU-S2 with ESP32S2. Added to v9.x only see here: https://github.com/adafruit/circuitpython/pull/8965


makermelissa (approved):
Thanks. Looks good.

PR #1334, opened 2024-02-25
Update Maker Uno RP2040 markdown file and images

noqman:
Hi Good day, here I updated the details and images for the Maker Uno RP2040 board.

Please let me know if there is any problem or questions. 

Thank you!


makermelissa (approved):
Thanks

PR #1337, opened 2024-03-01
Update adafruit_qtpy_esp32c3.md with references to the installation g…

jdimpson:
Added reference to https://learn.adafruit.com/circuitpython-with-esp32-quick-start/overview since the ESP32C3 chip requires different installation and workflow than is typical.


makermelissa (approved):
Thanks

PR #1338, opened 2024-03-01
Update seeed_xiao_esp32c3.md to refer to installation guide

jdimpson:
Add reference to https://learn.adafruit.com/circuitpython-with-esp32-quick-start/overview since ESP32C3 chips don't get installed or used the same was as typical cp devices.


makermelissa (approved):
Thanks, looks good.

PR #1340, opened 2024-03-06
Update waveshare_rp2040_touch_lcd_1_28.md

dglaude:
I hope this is not a breaking change, but currently there are two board that are "hard" to distinguish because they have the same name in the main download page:
* https://circuitpython.org/board/waveshare_rp2040_lcd_1_28/
* https://circuitpython.org/board/waveshare_rp2040_touch_lcd_1_28/


makermelissa (approved):
That should be fine. The board_id and filename are what can be breaking.

PR #1344, opened 2024-03-10
Update m5stack_dial documentation name in the link

dglaude:
Seems that this was made by copy and paste from another m5stack board.


makermelissa (approved):
Thanks.

PR #1345, opened 2024-03-11
Correct chip family for Teensy 4.0

snkYmkrct:
The Teensy 4.0 board was incorrectly tagged as part of the STM chip family.


dhalbert (approved):
Thank you!

PR #1350, opened 2024-03-17
Update espressif_esp32s3_devkitc_1_n8r8.md

vincemulhollon:
Added mfgr information for the RGB LED, mentioned which CircuitPython library is compatible with this LED, added information about the respin version 1.1 of the board which primarily moves the LED from GPIO 48 to GPIO 38

See https://docs.espressif.com/projects/esp-idf/en/stable/esp32s3/hw-reference/esp32s3/user-guide-devkitc-1.html#hardware-revision-details for details.

To quote Espressif:

"Both versions of ESP32-S3-DevKitC-1 are available on the market. The main difference lies in that the RGB LED is connected to different pins."

So its probably not worth the effort to create a 1.1 of the board that merely changes the alias for board.NEOPIXEL

### DIFF
--- a/_board/espressif_esp32s3_devkitc_1_n8r8.md
+++ b/_board/espressif_esp32s3_devkitc_1_n8r8.md
@@ -40,7 +40,7 @@ There are three mutually exclusive ways to provide power to the board:
 - **Boot Button**: Download button. Holding down **Boot** and then pressing **Reset** initiates Firmware Download mode for downloading firmware through the serial port.
 - **Reset Button**
 - **USB-to-UART Bridge:** Single USB-to-UART bridge chip provides transfer rates up to 3 Mbps.
-- **RGB LED**: Addressable RGB LED, driven by GPIO48.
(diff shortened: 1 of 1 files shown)


dhalbert:
Extra pin names could be added, like `NEOPIXEL_V10` and `NEOPIXEL_V11`.

dglaude:
> Extra pin names could be added, like `NEOPIXEL_V1_0` and `NEOPIXEL_V1_1`.

That works for python land, but don't you want to know where the NEOPIXEL is connected to permit the boot color code or other indication from the kernel?

makermelissa:
I had the same issue with the Arduino Nano ESP32 S3 and ended up creating separate board definitions for each. I used the "inverted status led" version to represent the less common board.

tannewt:
I'd just make a new board def. That will be clearest for folks. The boards are well labeled with the version.

vincemulhollon:
> I'd just make a new board def. That will be clearest for folks. The boards are well labeled with the version.

Yes, that is true the underside of my board is labeled as "v1.1" and this is a genuine board I received directly from Espressif.

I am now of the opinion that there is some value in creating a new v1.1 board as there will surely be a v1.2 or v2.0 in the future, and being able to use board.NEOPIXEL code unchanged on both versions is also of value to users.  Also, there is some value in that if a user is using "available" GPIO38 on a V1.0 board, GPIO38 is now "unavailable" on a V1.1 board and the user should be aware of that.

PR #1355, opened 2024-03-21
sort rss feed with newest at top

### LINKED ISSUE #1354: RSS Board Feed not showing most recent boards first
I would like to see the RSS feed in the top right corner of the CircuitPython.org website to be updated with the latest added boards. This means sorting the RSS feed in chronological order.  Most people expect an RSS feed to be in chronological order.  

The alphabetical RSS feed gives the appearance of not being updated since May 2023 if the expectation is that an RSS feed is in chronological order.  It gives the appearance of a dead RSS feed.

Capture

FoamyGuy:
resolves #1354 

After this change there were two devices showing up at the top of the list (out of chronological order) It turns out the date parsing format is very specific, it needs 4 digit years and zero padded single digits, with those changes these devices now get put into the proper place of the order.

Worth considering breaking out the Blinka boards as a separate feed. Right now they effectively get rendered one after another in the same giant RSS feed. This means they won't ever appear at the top of the feed for any RSS client that doesn't do it's own sorting locally on pubDate. Or another option is to try to intermix them into 1 list, and then sort that and render it as the RSS feed, I'm not certain if it's possible but can check into it if that is something that other folks think is a good way to do it.

### DIFF
--- a/_board/espressif_esp32s3_box_lite.md
+++ b/_board/espressif_esp32s3_box_lite.md
@@ -7,7 +7,7 @@ manufacturer: "Espressif"
 board_url:
  - "https://www.adafruit.com/product/5511"
 board_image: "espressif_esp32s3_box_lite.jpg"
-date_added: 22-07-08
+date_added: 2022-07-08
 family: esp32s3
 features:
   - Wi-Fi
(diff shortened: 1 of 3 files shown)


makermelissa (approved):
Looks good. The change of the date on the tdisplay shouldn't be necessary and there are probably many more boards without the leading 0.

PR #1356, opened 2024-03-21
Sort boards by date in the rss feed

jepler:
The fix in #1355 was incomplete, because it still places all blinka boards below all circuitpython boards. Also, we can do better at checking that dates are in a format that jekyll will be able to properly sort.

This also reformats all front matter due to how the file is processed. :-/ I could spend more time to make a more minimal change, if that's desirable.

The transformation was done with
```py
import datetime
import sys
import frontmatter

for i in sys.argv[1:]:
    post = frontmatter.load(i)
    date_added = post.get("date_added", "")

    if isinstance(date_added, str):
        try:
            date_added = datetime.datetime.strptime(date_added, "%Y-%m-%d").replace(hour=12)
        except ValueError as exc:
            print(f"Failed to parse date {date_added} in {i}: {exc}")
    elif isinstance(date_added, datetime.date):
        date_added = datetime.datetime.combine(date_added, datetime.time(hour=12))

    post["date_added"] = date_added

    with open(i, "wb") as f:
        frontmatter.dump(post, f)
        f.write(b"\n") # Re-add trailing newline
```


dhalbert:
Is this an alternative to #1355 or a continuation of that?

jepler:
The fix in https://github.com/adafruit/circuitpython-org/pull/1355 was incomplete, because it still places all blinka boards below all circuitpython boards. Also, we can do better at checking that dates are in a format that jekyll will be able to properly sort.

makermelissa:
Normally, the RSS reader handles the sorting and I'm certain this was the case when it was initially tested. The reason Blinka boards are placed separately from the CircuitPython boards is because they are under their own category via RSS.

jepler:
This PR does maintain the <category> tag while mixing the two lists togethre.

FoamyGuy:
I tested this out locally and it does look like it's working successfully to me. The contents of the local RSS are now intermixed and represent reverse chronological order of release. This does feel like it's most correct to my mind, I would be in favor of merging this change.

With that being said, practically speaking inside of "FeedR" (Android RSS reader app) it doesn't seem to make a difference. Both the local feed from this branch, and the currently live feed on the real site are sorted basically the same. It appears to be respecting the `pubDate` value and sorting on that. So even the original alphabetically sorted list was actually appearing chronologically inside the RSS reader app. In both cases Blinka and Circuitpython devices are intermixed into one list inside the app. I don't see any way within the UI of this app to access the category or sort by that, but you can still tell which devices are which by clicking or inspecting the link and seeing if it includes /board or /blinka


It does look there are other changes made to the formatting of the data files as well like removing quotes. If we do merge this we should probably update the guide page here: https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/adding-to-downloads to reflect the updated syntax standards.

makermelissa:
My main concern with this change is adding the time on there making it more difficult to maintain. Is there a way to do it without adding the 12:00 on there?

jepler:
I can try it without the 12:00s.

PR #1357, opened 2024-03-22
Fix board dates to always be of the form YYYY-MM-DD

jepler:
.. leading zeros are required for `frontmatter` to treat them as dates
rather than strings, apparently per the YAML specification.

This was done by script:
```py
import re
import datetime
import pathlib
import sys

import frontmatter

rx = re.compile(r'^(\s*)date_added:.*$', re.M)

for path_str in sys.argv[1:]:
    print(path_str)
    path = pathlib.Path(path_str)

    post = frontmatter.load(path)
    date_added = post.get("date_added", "")

    if isinstance(date_added, datetime.date):
        continue

    if isinstance(date_added, str):
        try:
            date_added = datetime.datetime.strptime(date_added, "%Y-%m-%d")
        except ValueError as exc:
            print(f"Failed to parse date {date_added} in {path_str}: {exc}")
            continue

    date_added = date_added.date()
    content = path.read_text("utf-8")
    new_content = rx.sub(lambda m: f"{m.group(1)}date_added: {date_added}", content)
    assert content != new_content

    path.write_text(new_content, "utf-8")
```

This PR replaces #1356 and I think is preferable because it doesn't make indentation & order changes to the .md files and it doesn't require adding an unused time-of-day to each board's date_added field.

### DIFF
--- a/_blinka/adafruit_feather_rp2040.md
+++ b/_blinka/adafruit_feather_rp2040.md
@@ -9,7 +9,7 @@ board_url:
 board_image: "adafruit_feather_rp2040.jpg"
 download_instructions: "https://learn.adafruit.com/circuitpython-libraries-on-any-computer-with-raspberry-pi-pico"
 blinka: true
-date_added: 2021-12-6
+date_added: 2021-12-06
 features:
   - Feather-Compatible
   - STEMMA QT/QWIIC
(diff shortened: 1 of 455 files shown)


makermelissa (approved):
Awesome. Thank you.

PR #1362, opened 2024-03-26
bump esp32s2 & esp32s3 bootloader version

felixerdy:
Bumping ESP32-S2 and ESP32-S3 bootloader versions.

Check https://github.com/adafruit/tinyuf2/releases/tag/0.18.2


makermelissa (approved):
Thanks.

PR #1375, opened 2024-03-29
Add missing boards from CircuitPython 9.1.0

### LINKED ISSUE #1374: Incorrect location of images when testing locally
It appears #1358 used an absolute path, which breaks the site for the purposes of local testing as it attempts to load non-existent images from circuitpython.org. Since the absolute path is only needed for RSS feeds, the URL should have been added to the feed page only.

makermelissa:
Also fixes #1374.


dhalbert (approved):
I am going to merge this to get the missing boards in for now. I am not sure what will happen for #1378, but if there's a merge conflict it can be redone.

jepler:
I can re-do 1378 if needed.

makermelissa:
> I can re-do 1378 if needed.

If you like @jepler. I just created a version of the image for feeds, so it gets around the problem. If you want to figure out how to get the into the same code block, that would be cool, but no hurry at this point.

PR #1378, opened 2024-03-30
Generate absolute URLs automatically

### LINKED ISSUE #1374: Incorrect location of images when testing locally

jepler:
This allows a `localhost:4000`-style URL to be generated for local testing, but still gives an absolute URL, which enhances compatibility with RSS readers. A comment is added explaining the reason for the absolute URL.

Closes #1374

### DIFF
--- a/_includes/downloads/board_image.html
+++ b/_includes/downloads/board_image.html
@@ -1,7 +1,8 @@
   {% assign small_image = "/assets/images/boards/small/" | append: include.board_image %}
   {% assign large_image = "/assets/images/boards/large/" | append: include.board_image %}
-  <img srcset="{{ small_image }} 300w,
-               {{ large_image }} 700w"
+  <!-- (uses an absolute URL for better compatibility with RSS readers -->
+  <img srcset="{{ small_image | absolute_url }} 300w,
+               {{ large_image | absolute_url }} 700w"
        sizes="(max-width: 1024px) 700px,
               300px"
        src="{{ large_image }}" alt="Image of Board" loading="lazy">


dhalbert (changes requested):
This now has merge conflicts due to merging #1375. But I am confused too: where does `absolute_url` come from?

dhalbert (approved):
OK, I see `absolute_url` is predefined, apparently. I'll approve this. Could you merge when you are ready to check the result immediately? Thanks.

jepler:
absolute_url: https://jekyllrb.com/docs/liquid/filters/#absolute-url

let's do this now if you're able

jepler:
generation looks good
```html
  <!-- (uses an absolute URL for better compatibility with RSS readers -->
  <img srcset="https://circuitpython.org/assets/images/boards/small/raspberry_pi_pico.jpg 300w,
               https://circuitpython.org/assets/images/boards/large/raspberry_pi_pico.jpg 700w" sizes="(max-width: 1024px) 700px,
              300px" src="/assets/images/boards/large/raspberry_pi_pico.jpg" alt="Image of Board" loading="lazy">

            ```

jepler:
hmm I lost the | absolute_url on the src=, oops

makermelissa:
> hmm I lost the | absolute_url on the src=, oops

You may need to point https://github.com/adafruit/circuitpython-org/blob/main/feed.html#L20 to board_image.html rather than feed_board_image.html.

PR #1379, opened 2024-03-30
One more absolute_url for rss feeds

jepler:


dhalbert (approved):

dhalbert:
@jepler Merge if you're ready

dhalbert:
Never mind, I will go ahead. I think you're offline.

PR #1383, opened 2024-04-01
Continue trying to fix RSS feed absolute URLs

jepler:
Originally, both the RSS feed and the board list pages used the same board_image.html include; I added `| absolute_url` to this file for the RSS feed, which functions better when all links are absolute.

Using `| absolute_url` is preferable to just prepending `https://circuitpython.org` because it can give a correct URL when previewing locally with `bundle run jekyll serve`.

This crossed with a change by @makermelissa to switch the feeds to use a separate include, presumably (though I didn't specifically ask her about it) to avoid having the absoulute URLs where they were not needed.

In this commit, I remove the use of `absolute_url` in the include that's used to generate regular pages, while switching from prepending `https://circuitpython.org` to using `| absolute_url` on the include that is used just for the feed.

### DIFF
--- a/_includes/downloads/board_image.html
+++ b/_includes/downloads/board_image.html
@@ -1,8 +1,7 @@
   {% assign small_image = "/assets/images/boards/small/" | append: include.board_image %}
   {% assign large_image = "/assets/images/boards/large/" | append: include.board_image %}
-  <!-- (uses an absolute URL for better compatibility with RSS readers -->
-  <img srcset="{{ small_image | absolute_url }} 300w,
-               {{ large_image | absolute_url }} 700w"
+  <img srcset="{{ small_image }} 300w,
+               {{ large_image }} 700w"
        sizes="(max-width: 1024px) 700px,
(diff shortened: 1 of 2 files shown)


jepler:
Tested locally, the download list page images work and are relative; the rss feed links are absolute links to localhost:4000.

makermelissa:
Do you know if the absolute_url works for loading images locally? If so, you could probably just combine the 2 pieces of code.

makermelissa:
> If you want me to re-do this PR so that it re-combines the two include files I'm happy to do that. My assumption had been that since you split them out subsequent to my two PRs adding `|absolute_url` the first time that either you were encountering a problem with the version I had written, or you didn't want the redundant URLs in view-source.

Please do. I was just in a hurry as I was at the end of my day and it was the simplest way to make the change that I knew would work for certain.

jepler:
Updated.

makermelissa (approved):
Looks good. Thanks.

PR #1384, opened 2024-04-02
Fix the feature list for the two Waveshare -GEEK boards

RetiredWizard:
@jvprat spotted some mistakes in the new Waveshare RP2040-GEEK and ESP32-S3-GEEK devices. The RP2040 device does not support WiFi and neither board has a USB-C port. While double checking the feature list I also realized that the board does have a Stemma QT compatible connector so I added that feature as well.


makermelissa (approved):
Thanks for the update.

PR #1388, opened 2024-04-06
Updated UM board info.

UnexpectedMaker:
- Fixed/Updated store URLs
- Updated info on Antennas
- Added more info for TinyWATCH S3


makermelissa (approved):
Thank you!

PR #1389, opened 2024-04-08
Update to peaceiris/actions-gh-pages@v4

dhalbert:
updates remaining action that needed update to Node.js 20. See https://github.com/peaceiris/actions-gh-pages/issues/1070.


tannewt (approved):
Thanks!

PR #1392, opened 2024-04-19
Correct name of ESP32-S3 HackTablet

dhalbert:
The title and name of this board was mistakenly "ESP32-S3 Box". I think this was copy-pasta.


makermelissa (approved):
Thanks

PR #1394, opened 2024-04-19
Update and add missing boards for latest CP release

makermelissa:


dhalbert (approved):
Thanks for the quick updates!

PR #1397, opened 2024-04-24
add documentation for newly added spotpear esp32c3 device

SeanTheITGuy:


makermelissa (changes requested):
Ok, #1399 fixed the build issues on our end, but it looks like you'll need to make some adjustments to this PR.

makermelissa (approved):
Thank you. Looks good.

PR #1414, opened 2024-05-23
Update waveshare_esp32_s3_zero.md

Timeline8:
Removed old data about the GRB LED on IO10 as they have fixed this. Schematic on their site shows the RGB LED connected to IO21 and I also confirmed that 

led = NeoPixel(NEOPIXEL, 1)
led[0] = (255,0,0)

makes the LED turn RED.

Also no idea how to put in this request, but the board module on this does not have a BUTTON however the board has a button on IO01. Can the CP module be updated to include a BUTTON?


makermelissa (approved):
Thanks

PR #1416, opened 2024-05-29
Add entry for Autosport Labs ESP32-CAN-X2 board

brentpicasso:
Adds entry for Autosport Labs ESP32-CAN-X2 board

https://wiki.autosportlabs.com/ESP32-CAN-X2


makermelissa (changes requested):
It appears to be failing because you have non-standard features under the features heading. We have it limited to just a few because it is used for filtering. You can move your board specific ones into the description.

brentpicasso:
@makermelissa  Hi, I'm trying to figure out the date_added field format.   I tried both YYYY-D-M and YYYY-M-D  - and both are failing CI.  

Currently it is date_added: 2024-28-5. 

How should I format the date_added field? 

Thanks!

jepler:
The date should be in YYYY-MM-DD form, with leading zeros if necessary. So: 2024-05-28

makermelissa (approved):
Thanks, looks good. Leading zeroes on the dates aren't required.

jepler:
[we're discussing over on #1417 whether the YYYY-MM-DD format is required; I believe that it is]

PR #1417, opened 2024-05-29
check-boards: show the correct format in the error message

jepler:
The message will now say e.g.,
```
2021-1-21 is an invalid date for example-board. The format must be YYYY-MM-DD, using leading zeros if necessary
```

### DIFF
--- a/tools/check-boards.py
+++ b/tools/check-boards.py
@@ -86,7 +86,7 @@ def verify_date_added(folder):
                 print(f"date_added field is missing for {board_id}")
                 valid = False
             elif not valid_date(date_added):
-                print(f"{date_added} is an invalid date for {board_id}")
+                print(f"{date_added} is an invalid date for {board_id}. The format must be YYYY-MM-DD, using leading zeros if necessary")
                 valid = False
     return valid
 


makermelissa:
The leading zeroes are actually optional, so the correct date format is either YYYY-M-D or YYYY-MM-DD.

jepler:
The [yaml timestamp type](https://yaml.org/type/timestamp.html) is documented as requiring a 4-digit year and 2-digit day/month. We previously made changes to ensure that dates are always written as YYYY-MM-DD, because of [how jekyll treats front matter](https://github.com/adafruit/circuitpython-org/pull/1357).

Having some dates in a different format was one of the problems with correctly ordering the RSS feed. We need to flag this as an error during our CI process, because when the incorrect format loads as a string, the ordering based on the date field is not what is intended.

Detecting the problem _but not showing what format was actually required_ [caused trouble for a contributor.](https://github.com/adafruit/circuitpython-org/pull/1416)

```py
>>> import yaml
>>> yaml.safe_load('[2020-1-1, 2020-01-01]')
['2020-1-1', datetime.date(2020, 1, 1)]
```

makermelissa (approved):
Fair enough. Looks like the checks are more strict than the JavaScript.

PR #1418, opened 2024-05-30
Fixed links to wiki for ESP32-CAN-X2

brentpicasso:
Fixed links to: https://github.com/autosportlabs/ESP32-CAN-X2


dhalbert (approved):
thanks!

PR #1420, opened 2024-06-06
Add ThingPulse Pendrive S3

squix78:


makermelissa:
It looks like this is failing some of the checks. I'd recommend moving your custom features into your description as bullet points.

squix78:
Thank you so much! I was hoping that this pull request would also allow people to flash the board directly from https://circuitpython.org/board/ThingPulse_PendriveS3/ How can I get that section to show up?

makermelissa:
> Thank you so much! I was hoping that this pull request would also allow people to flash the board directly from https://circuitpython.org/board/ThingPulse_PendriveS3/ How can I get that section to show up?

You need to have the board family and bootloader id, which it looks like you have. You'll need to add your board to https://github.com/adafruit/tinyuf2, which I don't see in the latest release, but maybe you added it already. And it looks like you'll need to wait for another release of circuitpython that includes your board.

PR #1421, opened 2024-06-07
Add rpga feather

skerr92:
Adds the RPGA Feather to circuit python-org. Tindie shop link will be active soon, we can wait to merge this until that's fully active if you'd like. I'm expecting approval any hour now.


skerr92:
And if I would have Waite 2 minutes, I would have gotten the email that the listing was approved. All links are good now, or should be anyway :)

makermelissa (approved):
Awesome. Thanks.

PR #1422, opened 2024-06-10
files for Sunton-ESP32-2432S032C

bablokb:
Assets and board-file for Sunton-ESP32-2432S032C.

Note that there is a pending PR for CircuitPython (https://github.com/adafruit/circuitpython/pull/9316) for this board as well.


makermelissa (approved):
Thank you.

PR #1426, opened 2024-06-20
Add Arduino Nano 33 BLE Rev2 board

applecuckoo:
Board def is already merged on the CircuitPython side, see https://github.com/adafruit/circuitpython/pull/9347.


makermelissa (approved):
Thank you. Looks good.

PR #1427, opened 2024-06-27
Archi board

EricAlmendras:
add new board Archi.


makermelissa (approved):
Thank you. Looks good.

PR #1436, opened 2024-07-22
Add latest round of missing boards

makermelissa:
This adds CircuitPython missing boards and Blinka missing boards.


_board/unexpectedmaker_feathers2_neo copy.md
tannewt, line 1: I suspect this file needs to be renamed so it doesn't include `copy`.

tannewt (changes requested):
One weird filename. Looks good otherwise. Thanks for adding these!

_board/unexpectedmaker_feathers2_neo copy.md
makermelissa: Nice catch. I'll take a closer look.

makermelissa: Fixed.

dhalbert (approved):

PR #1438, opened 2024-07-23
Changed some details in the names of the "Seeed Studio XIAO ESP32C3" and "Seeed Studio XIAO ESP32C6" boards.

Carla-Guo:
We noticed that we couldn't find these two boards by searching for "ESP32-C3" and "ESP32-C6".So it needed a change in the title of these two boards, changing "ESP32C3" to "ESP32-C3" and "ESP32C6" to "ESP32-C6". Thank you!

### DIFF
--- a/_board/seeed_xiao_esp32c3.md
+++ b/_board/seeed_xiao_esp32c3.md
@@ -1,8 +1,8 @@
 ---
 layout: download
 board_id: "seeed_xiao_esp32c3"
-title: "Seeed Studio XIAO ESP32C3 Download"
-name: "Seeed Studio XIAO ESP32C3"
+title: "Seeed Studio XIAO ESP32-C3 Download"
+name: "Seeed Studio XIAO ESP32-C3"
 manufacturer: "Seeed Studio"
 board_url:
  - "https://www.seeedstudio.com/Seeed-XIAO-ESP32C3-p-5431.html"
--- a/_board/seeed_xiao_esp32c6.md
+++ b/_board/seeed_xiao_esp32c6.md
@@ -1,8 +1,8 @@
 ---
 layout: download
 board_id: "seeed_xiao_esp32c6"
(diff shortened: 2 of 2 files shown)


dhalbert (changes requested):
We go by the manufacturer's name, and Seeed chooses not to include the hyphens. See https://www.seeedstudio.com/catalogsearch/result/?q=xiao . So I don't think we should change these names. There are other boards with similar non-canonical names, such as the "Lolin C3 Mini" or Unexpected Maker "TinyC6". We would not change the name of the "TinyC6 "to the "Tiny ESP32-C6".

So I don't think we should take this PR, sorry.

If you search for "C6", you can find the board. You can also click on "Filters" and choose "esp32c6" as the Processor family.

Carla-Guo:
I apologize for not making my personal information clear, I do in fact work for seeed, and this change was agreed upon!

Carla-Guo:
I apologize for not making my personal information clear, I do in fact work for Seeed Studio, and this change was agreed upon.
Best Regards,
Carla

dhalbert:
@Carla-Guo Thanks. Are you going to change the product names on your website too? Or is this only for getting the search to work on https://circuitpython.org ? If we had a different search mechanism that matched "ESP32-C3" with "ESP32C3", would you be requesting this change? I am concerned about the inconsistency between the official name and the name we are using here.

I think many people would just search for "C3" or "C6", and then the boards would be easily found.

Carla-Guo:
@dhalbert Thank you,
Yes, this is just for getting the search to work on https://circuitpython.org/, and if there is different search mechanism that matched "ESP32-C3" with "ESP32C3",we will not request this change, and if you can do that mechanism, it's much appreciated!

dhalbert:
We have merged #1439. Try searching for `esp32-c6` now in https://circuitpython.org/downloads. You should see the Seeed Studio board. You may need to do a shift-reload in the browser to get the new Javawscript.

PR #1439, opened 2024-07-23
match hyphenated or unhyphenated search terms

dhalbert:
- Addresses #1438 

Some boards uses a hyphenated name for the microcontroller chip or other feature, and some do not. Right now the search will give different results in the two cases. For instance, searching for "ESP32C6" yields different results than "ESP32-C6".

This change makes the match ignore hyphens, producing a more inclusive search.

We used to use a regexp, which would have made for an easy fix, but that was too slow (see #1328), so this is "hand-coded".

Tested locally.

@Carla-Guo FYI


makermelissa (approved):
Looks good. Thanks.

PR #1446, opened 2024-08-07
Add Waveshare RP2040 One to CircuitPython.org

johnnohj:
Submitted following successful merge [here](https://github.com/adafruit/circuitpython/pull/9484).


makermelissa (changes requested):
Please don't change template.md to add search filters that are specific to your board. Instead add your features under the description area.

makermelissa (approved):
Thank you. I had forgotten about the issue to add castellated pads, but also would like to mark all boards at the same time when adding.

johnnohj:
Absolutely makes sense - I'd like to help if that issue has stalled. Thank you for your review!

PR #1453, opened 2024-08-17
Turn libraries list on board pages into links

tyeth:
This turns the library list on board pages into clickable docs links.
It supports having class links like busio.SPI and frozen modules.
There were a few modules that don't have docs (_asyncio for example), and one frozen module (adafruit_connection_manager) that doesn't follow the expected path (it's just connectionmanager), but ignoring that it's working well.

Context from discord:
> I was going to add the links to the modules lists on boards pages (after imaginging they existed earlier 😁 ). I've got that sorted but the frozen modules are more tricky, mainly removing adafruit_ from the name and sticking after https://docs.circuitpython.org/projects/ works, e.g. https://docs.circuitpython.org/projects/display_text but not for connection_manager as it's page is without the underscore. Can RTD do aliases to fix that or do I need a more cunning plan?

Tested on the pico page (for normal modules), and pyportal (for frozen modules)

### DIFF
--- a/_includes/download/board.html
+++ b/_includes/download/board.html
@@ -147,13 +147,25 @@
     {% if version.modules %}
     <p>
         Built-in modules available:
-        <span class="download-modules {% if version.stable %}stable{% else %}unstable{% endif %}">{{ version.modules | join: ', ' }}</span>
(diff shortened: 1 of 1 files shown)


tannewt (approved):
Neat! Thank you!

PR #1459, opened 2024-08-23
Add missing boards for latest CircuitPython release

makermelissa:


dhalbert (approved):
Thank you for adding these so quickly!

PR #1460, opened 2024-08-23
workflows/build.yml: update to actions/checkout@v4

dhalbert:
Missed this in the previous PR to update to Node.20. This should get rid of the last warning we're seeing on the builds.


makermelissa (approved):
Looks good.

PR #1471, opened 2024-09-10
Update description and images for the VIDI X V1.1

lovrojakic:
Hi!

In this PR, I would like to change the description, images and resolve the board_id change from a corresponding PR which specifies that this firmware is for HW revision V1.1 of VIDI X board:
https://github.com/adafruit/circuitpython/pull/9613

Thank you!


makermelissa (changes requested):
Please use vidi_x for the filename as before so it matches the board ID in CircuitPython (see https://github.com/adafruit/circuitpython/tree/main/ports/espressif/boards). Otherwise it will show up as 2 boards.

lovrojakic:
Hi, it should be ok now. Thank you!

makermelissa (approved):
Looks good now, thank you.

PR #1482, opened 2024-09-24
Makerfabs_tft7: replace asterisk with 'x'

RetiredWizard:
The asterisk character in the screen resolution and interface listing gets interpreted as a formatting instruction messing up the output. This replaces the '*' with an 'x'.


jepler (approved):
Thanks. There may be a way to escape the `*` characters but this is a fine solution.

PR #1500, opened 2024-10-21
add seeed_studio_esp32s3_makedown_file and seeed_studio_esp32s3 image

Jasionf:


_board/seeed_xiao_esp32s3.md
```diff
@@ -0,0 +1,50 @@
+---
+layout: download
+board_id: "seeed_xiao_esp32s3"
```
FoamyGuy, line 3: We don't currently have a board that has the id `seeed_xiao_esp32s3` in the circuitpython core repo yet. So this listing will not be able to properly list its firmware download

FoamyGuy (changes requested):
The actions run failed due to missing images in the `large` and `small` folders.  This page covers the process of scaling images to the appropriate sizes for those two folders. https://learn.adafruit.com/how-to-add-a-new-board-to-the-circuitpython-org-website/preparing-the-images

_board/seeed_xiao_esp32s3.md
Jasionf: Our ID is called is seeed_xiao_esp3_2s3_sense?

Jasionf: Our ID is called seeed_esp32_xiao_s3_sense

Jasionf: Our ID is called seeed_esp32_xiao_s3_sense?

Jasionf: Our ID is called seeed_xiao_esp32_s3_sense?

FoamyGuy: @Jasionf we need to figure out some details about the sense version PID and come to a consensus on how best to include the non-sense version. There are some active discussions around this in discord and over in the core repo under the PR where the sense version was added to the core.  Once we have a clear path forward I will post an update here with more info.

FoamyGuy: We'd like to have this device added to the core first with a PR to https://github.com/adafruit/circuitpython that will require VID / PID codes which can be obtained from pid.codes. There is a page here: https://pid.codes/howto/ that details the process of submitting a PR to get codes allocated that way. 

When this device does get added to the core it will be under the board_id `seeed_xiao_esp32s3`, and then after that is merged in we finalize this one to get it added to the downloads page.

In the meantime, I think that the sense version of the firmware can work on the non-sense device. You'll just need to be wary of some of the pins not being easily accessible to hardware.

limengdu: Looks like some other peeps used to do a this: 
https://github.com/adafruit/circuitpython/issues/9507#issuecomment-2298119726
But I wasn't able to xiao_esp32s3 inside circuitpython, did I miss something?

dhalbert:
I have looked at the Seeed Xiao ESP32S3 basic board and the Sense board in more detail. The difference, as @deshipu noted, is that the Sense has the camera add on board.

There is a special very dense multi-pin connector on the base board for connecting the camera board. Because of that, the pins used for the camera are otherwise inaccessible.

Therefore, I would be inclined to make the base board be an alias of the sense board in circuitpython.org, and just have one build. Then we can avoid rounding up another USB PID. We can leave the name as including Sense, since it has camera capability.

@FoamyGuy does this make "sense" (oy) to you?

FoamyGuy:
@dhalbert yes that sounds good to me :+1:

dhalbert:
Per the comments above, the choice is to make this board an alias of the other board. It could have different photos.

makermelissa:
Since this hasn't been updated, but I believe all the necessary info/decisions are done, I should be able to update the code for this to make it an alias.

PR #1501, opened 2024-10-21
add espressif esp32p4 function ev

FoamyGuy:


_data/bootloaders.json
```diff
@@ -16,6 +16,7 @@
         "esp32c3": {},
         "esp32c6": {},
         "esp32h2": {},
+        "esp32p4": {},
```
FoamyGuy, line 19: This PR initially failed the actions check before adding `esp32p4` here in the bootloaders.json file.

I'm not certain if this is the appropriate solution or not, but it does make the actions pass. Please let me know if there is a different more correct way to handle it.

tannewt: This looks good to me. We don't have TinyUF2 for the P4 yet so this is fine.

tannewt (approved):
Thank you!

PR #1503, opened 2024-10-21
update board_id for circuitart zero s3

FoamyGuy:
This board_id didn't match the core repo which broke the download link for this board, and added a 2nd "unknown" instance to the downloads listing.
image

I tested this fix locally and it seems to resolve the issue, leaving only a single instance listed and the firmware download buttons work now


tannewt (approved):
Thank you! Good sleuthing!

Issue #1390, opened 2024-04-14
If board supports two circuitpython versions then sometimes ignores language drop down when downloading firmware


tyeth:
Noticed this a while ago but couldn't quite work out the problem, it turns out the first firmware panel on a download page works as expected, but the second ignores the language preference drop down and always gets en-US for en-GB / English (UK).

I've previously seen the web installer similarly download the wrong one and install it, I've seen it at least three times.
My normal route involves web installer, or bin download, UF2 only for upgrading existing running boards.

To recreate the download issue, goto Wiznet 5500 EVB, and choose English UK for the second firmware panel (less stable version) 9.1 firmware, choose English UK, and download UF2, then notice it got you en-US firmware instead.

I've checked the feather-s2 and it downloads the correct version, so something funny going on. 
Just trying to establish a more reliable recreation using web installer

tyeth:
The weird bit is sometimes I see US english in one box and UK english in another, but when recreating in an incognito window both my language dropdowns were prepopulated with UK.

![image](https://github.com/adafruit/circuitpython-org/assets/6692083/01f309e1-721f-4f48-b070-fb635802750b)

![image](https://github.com/adafruit/circuitpython-org/assets/6692083/0b5a99b8-82f6-4fd9-a9ac-b79467e09219)

tyeth:
These are the recent versions where I've ended up with en-US (occasionally I may have left at en-US so take this list with a pinch of salt)
![image](https://github.com/adafruit/circuitpython-org/assets/6692083/f0bfffec-44ca-4339-88c6-535666770708)
![image](https://github.com/adafruit/circuitpython-org/assets/6692083/ec9892bf-bbaf-4dce-86e7-73e66dd09ad3)

tyeth:
Looks like the second language drop down doesn't get set if the web installer button isn't found in the first firmware box:
![image](https://github.com/adafruit/circuitpython-org/assets/6692083/7dac6d6c-927b-40e6-82e3-cf0a3f31beed)

```
download.js:49 Uncaught TypeError: Cannot read properties of null (reading 'setAttribute')
    at download.js:49:50
    at Array.forEach (<anonymous>)
    at updateFileLinks (download.js:46:9)
    at findAndSetLocale (download.js:63:11)
    at download.js:73:11
    at NodeList.forEach (<anonymous>)
    at updateLanguageMenus (download.js:56:18)
    at HTMLDocument.<anonymous> (download.js:13:7)
```


[...]

PR #1509, opened 2024-10-25
fix language selection for beta download

FoamyGuy:
Resolves: #1390

The lack of installer button caused an exception to be thrown when attempting to access attributes of null. 

Since this occured in a for loop going over the download buttons when it failed on the first iteration it stopped subsequent iterations from ever happening. 

The solution proposed is to check for null before attempting to access properties of the element.

I confirmed the issue reported in the issue on the W5500-EVB-Pico both on the live site and in a local build.

I tested the fix successfully on the W5500-EVB-Pico in the local build.


tannewt (approved):
Thank you!

Issue #1268, opened 2023-10-26
[lolin_s2_mini.md] Add info about custom SPI/I2C pinouts.


bato3:
Because there no I2C/SPI labels on S2 Mini, there are used [def from Wemos MP:](https://github.com/adafruit/circuitpython/blob/2c47221cac7c9c3be74e145639117ee1823bc20e/ports/espressif/boards/lolin_s2_mini/mpconfigboard.h#L34-L39) 

-  SCL - GPIO35
-  SDA - GPIO33
-  SCK - GPIO7
-  MOSI - GPIO11
-  MISO - GPIO9

But in web there are present different pinout on images: https://www.studiopieters.nl/esp32-s2/ 

Please add a note, as there is currently no simple method to check this. (In my case it was easier to define other pins as I2C)

PR #1510, opened 2024-10-25
add note about lolin s2 mini SPI and I2C pins

FoamyGuy:
Resolves: #1268 

Added note about no labels on silkscreen and which pins are configured by default


tannewt (approved):
Thanks!

Issue #1087, opened 2022-11-02
Scrollable Checklists for filters


makermelissa:
The manufacturers list is starting to get a bit long on the filters page. Having a scrollable checklist would make it a bit more manageable. Additionally, if we wanted to be able to have modules as an option, this would allow it to exist without taking up a ton more room. It should be fairly easy to implement by placing the checkboxes inside a scrollable div with a maximum height.

PR #1512, opened 2024-10-25
Scrollable manufacturers

FoamyGuy:
Resolves: #1087 

With this change the list has limited height and can scroll. 
[Screencast from 2024-10-25 13-34-47.webm](https://github.com/user-attachments/assets/758c81cd-5d32-4d8a-8e0e-1a68364b4fa9)


dhalbert (approved):
:+1:

PR #1518, opened 2024-10-28
swap the .BIN and .UF2 download buttons

FoamyGuy:
Devices that have downloads for both will list the UF2 download above the BIN download with this change.

image


dhalbert (approved):
Thanks!

PR #1524, opened 2024-11-02
change the module links to purple tags and remove the seperating commas

FoamyGuy:
Mentioned recently on discord by @jepler

Trying out changing the style on the built-in and frozen modules to match the newly added feature "tags". Also added an inverse effect on hover to help further convey to the user that they are clickable (unlike the feature ones currently)

They also still retain their "pointer" mouse icon on hover.

In the screenshot `countio` was hovered, but my screenshot utility can't show the mouse. 
image

### DIFF
--- a/_includes/download/board.html
+++ b/_includes/download/board.html
@@ -161,7 +161,7 @@
           <a target="_blank" class="library-link" href="https://docs.circuitpython.org/en/latest/shared-bindings/{{ module_name }}">
           {% endif %}
             {{ module_name }}
-          </a>{% if module_name != version.modules[version.modules.size - 1] %}, {% endif %}
+          </a>
           {% endfor %}
         </span>
     </p>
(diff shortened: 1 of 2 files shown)


dhalbert:
@jwcooper @tylerdcooper What would you say about this styling?

dhalbert:
The purple tags are really "loud" visually. I think I would prefer leaving the current styling for modules as is (but fix the comma spacing), and changing the feature tags to look like the module tags. The feature tags could also be made clickable, so you get a search of all boards with that feature.

FoamyGuy:
Closing this one. #1525 provides a fix for the commas and spacing as well as switches the features over to be styled like the modules. 

@tylerdcooper all of the modules listed are included in the UF2 file. If the user downloads that UF2 file they will be able to use any of those modules without having to manually load anything else.

Issue #1528, opened 2024-11-12
Boards/Blinka without Download Instructions


jwcooper:
I find the download_instructions button to be incredibly useful, so here is a list of all boards without the link enabled. 

List requested by limor so we can backfill boards that already have good instructions but aren't linked here yet.

```
_board/01space_lcd042_esp32c3.md
_board/0xcb_helios.md
_board/42keebs_frood.md
_board/8086_commander.md
_board/8086_usb_interposer.md
_board/ADM_B_NRF52840_1.md
_board/Seeed_XIAO_nRF52840_Sense.md
_board/TG-Watch.md
_board/ThingPulse_PendriveS3.md
_board/adafruit_esp32s3_camera.md
_board/adafruit_feather_esp32_v2.md
_board/adafruit_feather_esp32c6_4mbflash_nopsram.md
_board/adafruit_floppsy_rp2040.md
_board/adafruit_funhouse.md
_board/adafruit_huzzah32_breakout.md
_board/adafruit_itsybitsy_esp32.md
_board/adafruit_itsybitsy_rp2040.md
_board/adafruit_kb2040.md
_board/adafruit_led_glasses_nrf52840.md
_board/adafruit_macropad_rp2040.md
_board/adafruit_magtag_2.9_grayscale.md
_board/adafruit_matrixportal_s3.md
_board/adafruit_metro_esp32s2.md
_board/adafruit_metro_esp32s3.md
_board/adafruit_metro_m7_1011_sd.md
_board/adafruit_metro_rp2040.md
_board/adafruit_metro_rp2350.md
_board/adafruit_neokey_trinkey_m0.md
_board/adafruit_pixel_trinkey_m0.md
_board/adafruit_proxlight_trinkey_m0.md
_board/adafruit_qt2040_trinkey.md
_board/adafruit_qtpy_esp32_pico.md
_board/adafruit_qtpy_esp32c3.md
_board/adafruit_qtpy_esp32s2.md
_board/adafruit_qtpy_esp32s3_4mbflash_2mbpsram.md
_board/adafruit_qtpy_esp32s3_nopsram.md
_board/adafruit_qtpy_rp2040.md
_board/adafruit_qualia_s3_rgb666.md
_board/adafruit_rotary_trinkey_m0.md
_board/adafruit_sht4x_trinkey_m0.md
_board/adafruit_slide_trinkey_m0.md
_board/adafruit_trrs_trinkey_m0.md
_board/adafruit_vindie_s2.md
[...]

Issue #644, opened 2021-03-02
"Unlist" in-dev/non-standard/discontinued boards


ajs256:
Some boards' downloads aren't really intended for the public and could be suppressed. These fall into 3 main categories:

**In Development:**
* [PyGamer Advance](https://circuitpython.org/board/pygamer_advance/)
* [PyBadge AirLift](https://circuitpython.org/board/pybadge_airlift/)
* [Metro M7 1011](https://circuitpython.org/board/metro_m7_1011/)
* [Feather RP2040](https://circuitpython.org/board/adafruit_feather_rp2040/)
* [Game Boy M4](https://circuitpython.org/board/gb_m4/)

**Non-Standard:** (These are non-standard hardware configurations)
* [Trinket M0 Haxpress](https://circuitpython.org/board/trinket_m0_haxpress/)
* [Feather M0 Supersized](https://circuitpython.org/board/feather_m0_supersized/)

**Discontinued:**
* [pIRkey](https://circuitpython.org/board/pirkey_m0/)
* [FeatherS2 Prelease](https://circuitpython.org/board/unexpectedmaker_feathers2_prerelease/)

and probably more I don't recognize!

These should be "unlisted" and hidden from the main boards page. They could be accessible with a direct link, or through a filter to show unlisted boards, a la Sparkfun: 
<img width="1312" alt="image" src="https://user-images.githubusercontent.com/67526318/109576039-e1b82180-7aa7-11eb-858e-ee56f2297267.png">

I think that these clog up the main boards list if someone is just trying to find a CP board that meets their needs.

makermelissa:
Thanks for the suggestion. The main purposes of having the boards are to allow people with these boards to be able to download the bootloader and showcase the boards that CircuitPython can run on. As you're probably aware we have filters to be able to find boards with certain features, but if we were to unlist or hide the boards, then it would defeat the main purposes. 

Perhaps a better idea would be to only show ones available to purchase, though I'm not sure how easy that would be to implement.

PR #1529, opened 2024-11-12
More howto install links

FoamyGuy:
@ladyada 

This is the rest of the Adafruit devices that I found install instructions guide pages for. 

A few I could not find a page for:
- edgebadge, it could link to pybadge but then it will link to a firmware with different name, though it would still work I think
- pirkey, has guide with pages about using circuitpython but not installing.

### DIFF
--- a/_board/adafruit_esp32s3_camera.md
+++ b/_board/adafruit_esp32s3_camera.md
@@ -10,6 +10,7 @@ board_image: "adafruit_esp32s3_camera.jpg"
 date_added: 2023-10-27
 family: esp32s3
 bootloader_id: adafruit_camera_esp32s3
+download_instructions: https://learn.adafruit.com/adafruit-memento-camera-board/install-circuitpython
 features:
   - Wi-Fi
   - Bluetooth/BTLE
--- a/_board/adafruit_feather_esp32_v2.md
+++ b/_board/adafruit_feather_esp32_v2.md
@@ -10,6 +10,7 @@ board_image: "adafruit_feather_esp32_v2.jpg"
 date_added: 2022-08-19
 family: esp32
 downloads_display: true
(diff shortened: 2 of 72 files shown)


ladyada:
@makermelissa

FoamyGuy:
After cross referencing with the list from #1528 and narrowing down to ones that have "Adafruit" as their manufacturer this is the list of devices that I did not find guide pages for: 

```
x feather_m0_express_crickit.md
x adafruit_floppsy_rp2040.md
x adafruit_huzzah32_breakout.md
x edgebadge.md
x metro_nrf52840_express.md
x adafruit_vindie_s2.md
x pygamer_advance.md
x feather_m0_adalogger.md
x circuitplayground_express_displayio.md
x adafruit_metro_rp2350.md
x feather_radiofruit_zigbee.md
x feather_m0_basic.md
x pybadge_airlift.md
x adafruit_qtpy_esp32c3.md
x monster_m4sk.md
x qtpy_m0_haxpress.md
x adafruit_feather_esp32c6_4mbflash_nopsram.md
x pirkey_m0.md
```

To my eye, they appear to fall roughly into a few buckets:

- Older devices from when circuitpython was still quite new e.g. pirkey, feather m0 basic
- New, no guide yet e.g. metro rp2350
- never was released as far as I can tell e.g. metro nrf52840, pybadge airlift
- variations of other builds that don't have a dedicated guide page that I could find e.g. CPX+DisplayIO, Edgebadge

dhalbert:
> * edgebadge, it could link to pybadge but then it will link to a firmware with different name, though it would still work I think

I'd say that page should be mirrored and should mention both firmware images. "If you have this ... if you have that. ..."

> * pirkey, has guide with pages about using circuitpython but not installing.

pirkey is discontinued and was barely ever sold, if at all. I think it could even be removed from circuitpython.org. See #644.

ladyada (approved):

PR #1530, opened 2024-11-12
cpx+displayio & edgebadge

FoamyGuy:
A few more to follow up on comments from #1529 

ty @makermelissa for the cpx + displayio link.

I've added the EdgeBadge firmware button as an option on the pybadge learn guide page as well and added a link to that for the edgebadge file here.


dhalbert (approved):
Tested links.

PR #1535, opened 2024-11-14
Added many new tutorial links + small fixes

makermelissa:
I went through all the board without tutorial links and checked if I could find them in under 30 seconds or so and added them if I could find them.


dhalbert (approved):
Great!!

FoamyGuy (approved):
Looks good to me. Thanks for adding all of these and doing that cleanup to the other bits!

PR #1537, opened 2024-11-18
Fix M5Paper family

dhalbert:
Family is `esp32s3`, should be `esp32`. Caused a spurious "Update bootloader" section on the board page.


FoamyGuy (approved):
Thank you!

Issue #1536, opened 2024-11-18
Searching "pico 2" doesn't easily find the pico2


FoamyGuy:
It finds many devices with "pico" or "2" in the name.

PR #1538, opened 2024-11-18
exact match tags search

FoamyGuy:
Resolves: #1536 

This changes makes it so that exact matches on tags will get higher priority in the resulting list of boards. It adds "pico 2" to the pico 2 tags so that it will appear as the first result for "pico 2" now.


dhalbert (approved):
Fantastic!

PR #1543, opened 2024-11-24
Support Pimoroni Pico Plus 2W

eightycc:
Support for Pimoroni Pico Plus 2W. Complements adafruit/circuitpython#9826.


FoamyGuy (approved):
Thank you!

PR #1545, opened 2024-11-25
T-Deck (Plus)

bill88t:
Update the LILYGO T-Deck page to indicate support for it's plus variant.
The image now features the Plus on the left view.


FoamyGuy (approved):
Thank you!

PR #1551, opened 2024-12-02
add Music Thing Modular Workshop Computer info

todbot:
New `mtm_computer` board just got merged to CircuitPython. Here's the circuitpython.org info for it.


FoamyGuy (approved):
Thank you! I updated the family to `rp2040`. The `raspberrypi` family is now split up into the different processors, but the PR that made that change came it essentially the same time as this one.

PR #1552, opened 2024-12-04
split raspberrypi family to rp2040 and rp2350

FoamyGuy:
As discussed during a recent meeting. Split `raspberrypi` processor family into `rp2040` and `rp2350`


makermelissa (approved):
Looks good to me.

Issue #1556, opened 2024-12-28
Precheck on builds for circuitpython.org hangs - Wrong Ruby version


oschettler:
The precheck for [my pull request](https://github.com/adafruit/circuitpython-org/pull/1555) crashes:

````
bundle install
/opt/hostedtoolcache/Ruby/3.4.1/x64/bin/bundle config --local path /home/runner/work/circuitpython-org/circuitpython-org/vendor/bundle
/opt/hostedtoolcache/Ruby/3.4.1/x64/bin/bundle config --local deployment true
Cache key: setup-ruby-bundler-cache-v6-ubuntu-22.04-x64-ruby-3.4.1-wd-/home/runner/work/circuitpython-org/circuitpython-org-with--without--only--Gemfile.lock-1b183b6e22bb03317bf80c0c78c15ec3d7a18371af3deae5524f221991e14296
/opt/hostedtoolcache/Ruby/3.4.1/x64/bin/bundle install --jobs 4
Fetching gem metadata from https://rubygems.org/..........
google-protobuf-4.27.5-x86_64-linux requires ruby version < 3.4.dev, >= 3.0,
which is incompatible with the current version, ruby 3.4.1p0
Error: The process '/opt/hostedtoolcache/Ruby/3.4.1/x64/bin/bundle' failed with exit code 5
````

Here is a screenshot:

![IMG_3303](https://github.com/user-attachments/assets/95ad5b65-99c0-4710-8729-082875f1caa1)

PR #1557, opened 2024-12-28
update bundle

jepler:
Closes: #1556


jepler:
If we're going to pin every ruby dependency, should we also at least pin ruby to a minor version like `ruby-version: '3.4'`  rather than a major version like `ruby-version: '3'` during the build?

jepler:
at the moment the latest https://rubygems.org/gems/google-protobuf/versions/4.29.2-x86_64-linux does not work with 3.4.

PR #1562, opened 2025-01-03
Sunton_esp32_8048S050: update board info with CIRCUITPY_DISPLAY_FREQU…

RetiredWizard:
This will update the Sunton_esp32_8048S050 board information with the settings.toml parameter which was added in PR adafruit/circuitpython#9911


tannewt (changes requested):
Please include a note about what version it is available in. 9.2.2+ I think.

RetiredWizard:
Ah, good point! The update's not in stable yet :grin: Thanks!!!

tannewt (approved):
Great! Thanks!

PR #1563, opened 2025-01-03
Use the latest firmware installer package which uses esptool-js

makermelissa:
Tested with ESP32, ESP32-S2, ESP32-S3, and ESP32-C6. See https://github.com/adafruit/web-firmware-installer-js/pull/11 for more details.


dhalbert (approved):
👍

PR #1568, opened 2025-01-13
Add PicoMo V3

supcik:
I made a new firmware for the PicomoV3 and this is the corresponding doc. I also added "V2" ti the name of the version 2 to avoid confusion.


FoamyGuy (approved):
Thank you!

PR #1577, opened 2025-01-18
Fix: TypeError while searching on Blinka page

michalpokusa:
Typing any phrase in search bar on  "Blinka" page and then removing it does not correctly display all the cards.
The underlying issue is using `.split()` on `undefined`, which prevents the function from completing.

image

Spotted by @FoamyGuy during th Saturday stream: https://www.youtube.com/watch?v=8czmvgN3Q70&t=360s


FoamyGuy (approved):
This looks good to me. I tested successfully with a local build and confirmed it solves that problem. 

Thank you!

PR #1578, opened 2025-01-23
include how much flash in 8MB no psram ESP32-S3 board names

dhalbert:
Two ESP32-S3 boards did not include "8MB Flash" in their names. Do this explicitly.

Example of user confusion: https://discord.com/channels/327254708534116352/537365702651150357/1332080125835083839


FoamyGuy (approved):
Thank you!

PR #1579, opened 2025-01-24
Fix family for spotpear_esp32c3_lcd_1_44.jpg

dhalbert:
User had trouble using circuitpython.org installer: https://forums.adafruit.com/viewtopic.php?t=216227.
Turns out chip family designator was `esp32` instead of `esp32c3`.


FoamyGuy (approved):
Thank you!

PR #1580, opened 2025-01-27
Update arduino_nano_rp2040_connect.md

Timeline8:
Web URL was a "404" dead link. Updating it to the latest valid URL.


FoamyGuy (approved):
Thanks for the fix!

FoamyGuy (changes requested):
@Timeline8 could you make the same change up in the "board_url" section of the same file that you edited? It's got the same old / non-working URL up there.  (Sorry, I didn't think to check the rest of the file until after I clicked approve).

FoamyGuy (approved):

PR #1583, opened 2025-01-29
Add missing boards up to CP 9.2.4

makermelissa:
I also updated my missing boards script to include extra boards to help easily find things like misnamed board files such as with the cezerio_dev_ESP32C6.


_board/cezerio_dev_esp32c6.md
```diff
@@ -1,6 +1,6 @@
 ---
 layout: download
-board_id: "cezerio_dev_esp32c6"
+board_id: "cezerio_dev_ESP32C6"
```
dhalbert, line 3: It's too bad these are not lowercase, but someday maybe we can fix that.

dhalbert (approved):
Thanks!!

PR #1584, opened 2025-01-29
Make Feather ESP32 V2 show up in search for "huzzah"

jepler:
.. since this word is associated with the Feather ESP32 and is shown in large letters on the bottom of the V2 board.
image


makermelissa (approved):
Looks good to me. Thanks.

PR #1587, opened 2025-02-08
drop 8.x bundles

dhalbert:
As discussed in the weekly CircuitPython meeting, https://github.com/adafruit/adafruit-circuitpython-weekly-meeting/blob/main/2025/2025-02-03.md#2643-in-the-weeds, we will drop building 8.x bundles.

- This is on the checklist here: https://github.com/adafruit/circuitpython/issues/9001


jepler (approved):
Thanks!

PR #1592, opened 2025-02-24
Add HXR SAO Digital Multimeter board

flummer:


flummer:
Is there any more needed from my side?

If possible, it would be really awesome, if this could be merged ahead of Hackaday Berlin (happening March 14th to 16th 2025), as attendees will receive this device in their goodie bags (though we will preprogram them).

_board/hxr_sao_dmm.md
```diff
@@ -0,0 +1,57 @@
...
+name: "SAO Digital Multimeter"
+manufacturer: "HXR.DK"
+board_url:
+ - "https://github.com/flummer/dmm-sao"
+board_image: "hxr_sao_dmm.jpg"
+date_added: 2025-02-24
+family: rp2040
+downloads_display: true
+blinka: false
+download_instructions: ""
```
tyeth, line 14: Maybe stick in a link to the paragraph above the flashing instructions, where it mentions using a custom build of circuitpython (to get the needed libraries)

I'd probably link to:
https://github.com/flummer/dmm-sao/?tab=readme-ov-file#firmware

tyeth (approved):
This looks good, although probably worth adding a link to the flashing instructions under the `download_instructions` key, like this board does: https://github.com/adafruit/circuitpython-org/blob/main/_board/aramcon2_badge.md?plain=1#L15
(See earlier comment)

tyeth (approved):

PR #1597, opened 2025-03-01
Update bootloaders.json - tinyUF2 0.21.0

tyeth:
I hadn't spotted the bootloader version was a defined value, and since adding the TDisplayS3 bootloader option it needs the latest tinyUF2 release to pick it up.

Otherwise currently stalls at downloading bootloader stage:
image

This should pick up https://github.com/adafruit/tinyuf2/releases/tag/0.21.0


FoamyGuy (approved):
Thank you

PR #1617, opened 2025-04-09
Add new filter to only use featured guides from circuitpython category

jwcooper:
Should help curate the home page a bit better so we can limit guides with blurry images or guides that are out of date, etc.


makermelissa (approved):
Looks good to me.

PR #1632, opened 2025-05-12
Switch to download counts from Cloud Front

tannewt:
Previous numbers were from S3 direct logs. Cloud front requests from S3 on cache miss only. The user agent was also cloud front instead of the actual requester. Cloud front logs have the actual user agent and we use it to filter out bot requests.

So, the old numbers (relatively) under counted popular boards and over counted boards that only bots cared about. These numbers should be truer to reality.

### DIFF
--- a/_data/files.json
+++ b/_data/files.json
@@ -1,6 +1,6 @@
 [
  {
-  "downloads": 3149,
+  "downloads": 6,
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
Great! Should be more realistic.

PR #1636, opened 2025-05-15
Add new Blinka Boards

makermelissa:
This adds 16 new Blinka boards putting us at 162. This also updates the U2IF boards to clarify how they are running.


dhalbert (approved):
Looks good!

PR #1640, opened 2025-05-19
differentiate original Feather HUZZAH32 and Feather ESP32 V2

dhalbert:
A few users have confused the original Feather HUZZAH32 and Fthe eather ESP32 V2. Add some cautionary text. The latest user with this issue also asked that the name include "HUZZA32" (which is on the bottom of the board).

I also added some alert boxes already in the two guides.

I deliberately took the link to the original HUZZAH32 Feather out of the description of the V2. I also removed some obsolete "it's early" language from the original description.


TheKitty (approved):
looks good

PR #1659, opened 2025-07-15
Update espressif bootloader; update instrutions

dhalbert:
For CircuitPython 10, the ESP32-S2 and ESP32-S3 flash partition layouts have changed, to increase the size of the firmware partition. These changes are upward-compatible with CircuitPython 9.1.0 and up.  All the builds have now been updated to assume the new partition layout.

- Bump the TinyUF2 version for `esp32s2` and `esp32s3` to 0.3.50.
- Update and rewrite the Espressif bootloader install/update section, to make it clearer and point to external references.

Here is a screenshot of the revised section, for easier proofing:
image

### DIFF
--- a/_includes/download/board.html
+++ b/_includes/download/board.html
@@ -39,8 +39,8 @@
     {% endif %}
     {% if page.family == 'nrf52840' %}
       <p>
-        <b>On nRF boards, CircuitPython 8.2.0 and later require UF2 bootloader version 0.6.1 or later.
-        Older bootloaders cannot load the firmware. See <i>Update UF2 Bootloader</i> below.</b>
+        <strong>On nRF boards, CircuitPython 8.2.0 and later require UF2 bootloader version 0.6.1 or later.
+        Older bootloaders cannot load the firmware. See <i>Update UF2 Bootloader</i> below.</strong>
       </p>
     {% endif %}
     <p>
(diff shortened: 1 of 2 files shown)


_includes/download/board.html
```diff
@@ -227,80 +227,86 @@ <h3>Previous Versions of CircuitPython</h3>
...
+    The bootloader itself is not CircuitPython.
+    On Espressif boards, the UF2 bootloader is called <em>TinyUF2</em>.
+    If the TinyUF2 bootloader is installed, you can check its version by looking in
+    the <strong>INFO_UF2.TXT</strong> file when the <strong>BOOT</strong> drive is visible
+    (<strong>FTHRS2BOOT</strong>, <strong>MAGTAGBOOT</strong>, <strong>HOUSEBOOT</strong>, etc.)
   </p>
   <p>
-    It is not necessary to reinstall a UF2 bootloader you unless a <b>BOOT</b> drive is not visible
-    when in UF2 bootloader mode, or you know of a problem with your current UF2 bootloader.
+    In general, it is not in general necessary to update TinyUF2 at every version change.
```
FoamyGuy, line 246: small typo here. Could be "In general, it is not in generally necessary"

or just take out "generally" all together since it starts with "in general,"

FoamyGuy: Ah, and I realize after posting this that I mis-read it, and generally is wrong. 

But it could still be removed to avoid repetition if we want. I don't have a strong opinion either way now that I re-read it correctly.

FoamyGuy (approved):
Looks good to me.

PR #1677, opened 2025-08-25
fix crowpanel 3.5 images

FoamyGuy:
As noted by a user on discord:

> FYI, on the Circuitpython Downloads page for the Elecrow 3.5" touch LCD, https://circuitpython.org/board/elecrow_crowpanel_3.5/, the image is completely wrong - it's showing an epaper display iwth a case, not the 3.5" board with a touch LCD.

It turns out we did already have a crowpanel 3.5 photo, but the .md file had been pointing to the wrong image. But I did swap it to a new image that shows the LCD rather than the back of the board along with correcting the the md file.


dhalbert (approved):
Thanks!

PR #1680, opened 2025-09-04
Add Ai-On-The-Edge-Cam board

allexoK:


allexoK:
Hmm, I don't understand why the check fails...
Annotaions section suggest that there is some problem with 'features', I also see 'could not find expected ':''. So I guess it couldn't find ':', but it's in place.

allexoK:
Wow, putting s space between the 'board_url:' and the actual link worked!

makermelissa (approved):
Thank you.

PR #1698, opened 2025-10-18
Add missing CP 10.0.3 boards and new Blinka board

makermelissa:
This adds any missing boards for CircuitPython as of 10.0.3 and a missing Blinka board.


dhalbert (approved):
Thanks!

PR #1753, opened 2026-04-14
added Pimoroni Badger2350

bablokb:
Support for Pimoroni Badger2350 was added with https://github.com/adafruit/circuitpython/pull/10929


makermelissa (approved):
Looks good. Thank you!

PR #1754, opened 2026-04-14
fix paths to images

bablokb:
The target directories for images were missing the `/boards/` directory component.


FoamyGuy (approved):
Thanks for the fix!

PR #1759, opened 2026-04-16
update GitHub CI action versions for Node.24

dhalbert:
Updated `actions/checkout`, `actions/setup-python` to latest, for Node.24.

Also updated `peter-evans/create-pull-request` from `@v4` to `@v8`. This is a big jump. Let's see if it works. (EDIT: will not get tested until the image updater job runs).


makermelissa (approved):
Thanks. Looks good and it seems to be passing the checks when you submitted this.

Issue #1760, opened 2026-04-21
Sorting boards by date isn't working properly.


makermelissa:
When sorting boards on the downloads page, if you sort by alphabetical, it sorts fine, but when trying to sort by date (particularly newest first), it give the boards in an unexpected order.

PR #1761, opened 2026-04-21
Fix board date sorting on downloads page

makermelissa-piclaw:
## Problem

Sorting boards on the downloads page by date (newest/oldest first) produces unexpected results, as reported in #1760.

## Root Cause

The sort comparator in `downloads.js` had several issues:

1. **Tag match overrides all sort modes**: The exact tag match promotion (`return -2`) ran before the sort-type switch statement for *all* sort modes, so when a search term was active, matching boards were always pushed to the top — even when the user explicitly selected date or alphabetical sorting.

2. **Asymmetric tag comparison**: Only item `a` was checked for tag match; item `b` was never tested. This made the comparator inconsistent (violating the sort contract where `compare(a,b)` and `compare(b,a)` should be symmetric).

3. **Missing equality case in comparators**: The date and download-count comparators used ternary returns (`< ? -1 : 1`) instead of subtraction, so they never returned `0` for equal values. This produced unstable sort results for boards sharing the same date or download count.

4. **filterResults also ignores sort selection**: The `filterResults` function prepends tag-matched boards to the DOM regardless of the currently selected sort, undoing an explicit date or alphabetical sort whenever filtering re-runs.

## Fix

- Tag match promotion now only applies when the sort type is `downloads` (the default)
- Both `a` and `b` are checked for tag matches symmetrically
- Date and download comparators use arithmetic subtraction for proper three-way comparison
- `filterResults` checks the active sort radio before doing tag-match reordering

## Testing

- Sort by "Date Added (Newest First)" → boards appear in correct descending date order
- Sort by "Date Added (Oldest First)" → boards appear in correct ascending date order  
- Sort by alphabetical → unaffected by search tag matches
- Default downloads sort with search term → tag matches still promoted to top (existing behavior preserved)

### DIFF
--- a/assets/javascript/downloads.js
+++ b/assets/javascript/downloads.js
@@ -350,8 +350,10 @@ function filterResults() {
     } else {
       download.style.display = 'block';
       board_count++;
-      // exact tag match re-order
-      if (downloadsSearch.searchTerm !== null && downloadsSearch.searchTerm !== undefined) {
+      // exact tag match re-order only for default (downloads) sort
+      var currentSort = document.querySelector("input[name='sort-by']:checked");
+      var currentSortValue = currentSort ? currentSort.value : 'downloads';
+      if (currentSortValue === 'downloads' && downloadsSearch.searchTerm !== null && downloadsSearch.searchTerm !== undefined) {
         let searched = downloadsSearch.searchTerm.toLowerCase();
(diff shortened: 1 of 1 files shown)


makermelissa:
Tested locally and it is behaving as expected. Actually, a bit better because new boards are now at the end of the list.

makermelissa (approved):

PR #1777, opened 2026-05-17
update samx bootloaders to v4.0.0

dhalbert:
Update the SAMD and SAME bootloaders to v4.0.0, which includes a vital fix https://github.com/adafruit/uf2-samdx1/pull/231 to allow BOOT drives to work on ChromeOS and recent versions of Linux that use `fwupd` 2.x.x.


makermelissa (approved):
Thank you

PR #1779, opened 2026-05-27
Update purchase link and add hardware info

Timeline8:
Updated purchase link and added hardware information.
Purchase link gave me (in the USA) a 404 page. So updated to one that worked. Also found the Github page for that specific board and added the link.

### DIFF
--- a/_board/01space_lcd042_esp32c3.md
+++ b/_board/01space_lcd042_esp32c3.md
@@ -41,4 +41,7 @@ With its state-of-the-art power and RF performance, this SoC is an ideal choice 
 
 ## Purchase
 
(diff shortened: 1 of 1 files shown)


dhalbert (approved):
These Aliexpress links are rather ephemeral, so if it changes again, maybe we can just add a description to search for.

Timeline8:
Yeah I agree that unless it is a known manufacturer with a stable Aliexpress storefront (for example Wemos Lolin, WeAct, etc) a description might be a better option. At least I found the github links for this one so that should be description enough if the Ali link I added dies next.

Thanks!

PR #1787, opened 2026-06-25
Update purchase link for Nano RP2040 Connect

Timeline8:
The "With Headers" version link is dead and that version does not appear to be available any longer. Updating the URL to reflect the current (no header) version available in the Arduino store.


makermelissa (approved):
Thank you

PR #1789, opened 2026-07-07
Update Franzininho WiFi Wroom board images

FBSeletronica:
Replace outdated Dev Edition photos with current production Franzininho WiFi Wroom board images (original, large, small).


dhalbert (approved):
Thank you!

PR #1794, opened 2026-07-24
Rename yd_esp32_s3_n16r8.md to YD_esp32_s3_n16r8.md

seleste3580-dev:


dhalbert:
We use all lower case for board names. There are a few exceptions but they are from before we set the rule. In this case, the board is already called `yd_esp32_s3_n16r8` in the `circuitpython` repo.

PR #1796, opened 2026-07-29
add te ep2350

FoamyGuy:
Adding Teenage Engineering EP-2350 ting.

This device was added to the core in https://github.com/adafruit/circuitpython/pull/11136


makermelissa (approved):
Looks good. Thanks.

PR #1799, opened 2026-08-12
Add ESP32-S3-DevKitC-1-N32R16, fix N32R8 product link

mikeysklar:
Adds the download page for `espressif_esp32s3_devkitc_1_n32r16`, the board def added in adafruit/circuitpython#11192. Verified on hardware by @grgrant: 32MB flash, 16MB PSRAM, NeoPixel on GPIO38.

Two things about the existing `n32r8` page, both in the second commit:

- It links [Adafruit 5364](https://www.adafruit.com/product/5364) as the place to buy it, but 5364 is the **N32R16V** (32MB flash, 16MB PSRAM). Adafruit does not sell an N32R8V at all. Someone buying 5364 and searching the downloads page currently lands on the N32R8 entry, which is what prompted adafruit/circuitpython#11190.
- The N32R8V module is End of Life per Table 1-1 of the [WROOM-2 datasheet](https://www.espressif.com/sites/default/files/documentation/esp32-s3-wroom-2_datasheet_en.pdf), marked EOL in rev 1.4 (2024-12-03).

I removed the incorrect `board_url` and replaced the Purchase section with an Availability note pointing at the new page. Happy to restore a purchase link if there's a correct target I missed.

I did **not** touch the "RGB LED, driven by GPIO48" line on the n32r8 page. GPIO48 is correct for the v1.0 board that def targets. The v1.0/v1.1 LED split is adafruit/circuitpython#9275.

I also left `bootloader_id` off the new n32r16 page deliberately. The only `espressif_esp32s3_devkitc_1` tinyuf2 build is configured for 8MB flash, and @grgrant measured CIRCUITPY dropping to 5,935,104 bytes under it versus 31,080,448 flashed with esptool. Tracked in adafruit/tinyuf2#481.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/_board/espressif_esp32s3_devkitc_1_n32r16.md
+++ b/_board/espressif_esp32s3_devkitc_1_n32r16.md
@@ -1,0 +1,45 @@
+---
+layout: download
+board_id: "espressif_esp32s3_devkitc_1_n32r16"
+title: "ESP32-S3-DevKitC-1-N32R16 Download"
+name: "ESP32-S3-DevKitC-1-N32R16"
+manufacturer: "Espressif"
+board_url:
+ - "https://www.adafruit.com/product/5364"
+board_image: "espressif_esp32s3_devkitc_1.jpg"
+date_added: 2026-08-12
+family: esp32s3
+features:
+  - Wi-Fi
+  - Bluetooth/BTLE
+  - Breadboard-Friendly
+---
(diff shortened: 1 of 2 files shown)


makermelissa (approved):
Thank you

PR #1806, opened 2026-09-03
add te sp1 to downloads  page

FoamyGuy:
board def is merged in core: https://github.com/adafruit/circuitpython/pull/11261


makermelissa (approved):
Thank you, looks great.

PR #1807, opened 2026-09-08
Add Challenger RP2040 NFC

TheFilipcom4607:
The board definition was merged in adafruit/circuitpython#11321, so `challenger_rp2040_nfc` builds and already has nightlies on S3, but it has no page here. It was the only Challenger RP2040 variant missing from the downloads list.

Adds:

* `_board/challenger_rp2040_nfc.md`
* `assets/images/boards/{original,large,small}/challenger_rp2040_nfc.jpg` at 1365x1024, 800x600 and 300x225

A few notes on choices I made, all easy to change:

* The photo is iLabs' own product image from their product page, the same source as the other Challenger listings. I am not the manufacturer, just someone who owns the board and added the definition upstream.
* `download_instructions` is left out on purpose. The sibling Challenger pages point it at the "Getting started" tab on ilabs.se, but that tab covers the Arduino core only, and the field renders as "Learn how to install CircuitPython on this board".
* The page spends a few lines on the PN7150 because two things about it otherwise look like a broken board: its I2C bus needs 100 kHz, since the bus has no external pull-ups and the board relies on `CIRCUITPY_I2C_ALLOW_INTERNAL_PULL_UP`, and `i2c.scan()` never lists the chip because it ignores zero-length writes. Happy to trim that section if you would rather keep the page short.

`tools/check-boards.py` and `tools/check-images.py` both pass locally.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


TheFilipcom4607:
Pushed a small update: the driver was just merged into the Community Bundle (adafruit/CircuitPython_Community_Bundle#288), so the install line is now plain `circup install pn7150` instead of the `circup bundle-add` step it needed before.

makermelissa (approved):
Thanks, looks good.

PR #1808, opened 2026-09-14
Device info and image assets for the Seeedstudio XIAO ESP32 S3 TRMNL Kit

bablokb:
The device has been added to CP already.


tannewt (approved):
Thank you!

PR #1813, opened 2026-09-17
fixing inconsistent board_id and filename for xiaomao

securelyfitz:
Oops - i made a mistake with the board_id in my last pull request. Fixing this to be consistent with what's in the circuitpython repository so it all lines up. Apologies!


makermelissa (approved):
Thanks for updating

PR #1815, opened 2026-09-17
blinka manufacturer links filter blinka not circuitpython downloads

securelyfitz:
I noticed that if you click a manufacturer name on a blinka page, it links you to the circuitpython downloads for that manufacturer, not blinka devices. 
For example https://circuitpython.org/downloads?manufacturers=Adafruit instead of https://circuitpython.org/blinka?manufacturers=Adafruit 

makes more sense to me but perhaps intentional for other reasons.


makermelissa (approved):
Thanks

